PR #15, opened 2025-10-01
Load user details on app initialization

shayanmalinda:


apps/people-app/webapp/src/route.ts
```diff
@@ -53,21 +53,21 @@ export const routes: RouteObjectWithRole[] = [
...
-    allowRoles: [Role.ADMIN, Role.TEAM],
+    allowRoles: [Role.ADMIN],
   },
   {
     path: "/profile",
     text: "Profile",
     icon: React.createElement(PersonIcon),
     element: React.createElement(View.profile),
-    allowRoles: [Role.ADMIN, Role.TEAM],
+    allowRoles: [Role.ADMIN],
```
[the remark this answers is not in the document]
shayanmalinda: These are just sample routes. Routes will be configured later on.

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -15,6 +15,21 @@
...
+isolated function getEmployeeBasicInfoQuery(string email) returns sql:ParameterizedQuery =>
+    ` SELECT 
+            e.id,
+            e.first_name,
+            e.last_name,
+            e.wso2_email,
+            e.employee_thumbnail,
+            e.job_role
+        FROM employee e
+        WHERE e.wso2_email = ${email} LIMIT 1;`;
```
Rashmika998, line 31:
```suggestion
isolated function getEmployeeBasicInfoQuery(string email) returns sql:ParameterizedQuery => `
    SELECT 
        id,
        first_name,
        last_name,
        wso2_email,
        employee_thumbnail,
        job_role
    FROM employee 
    WHERE wso2_email = ${email}`;
```
Since `queryRow()` method is being used, we don't need to specify `LIMIT 1`

apps/people-app/backend/service.bal
```diff
@@ -51,6 +51,43 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+        if employeeBasicInfo is error {
+            string customError = string `Error occurred while fetching employee information`;
+            log:printError(customError, employeeBasicInfo, email = userInfo.email);
+            return <http:InternalServerError>{
+                body: {
+                    message: customError
+                }
+            };
+        } else if employeeBasicInfo is () {
+            log:printWarn("No employee information found for the user: " + userInfo.email);
```
Rashmika998, line 79:
```suggestion
            log:printWarn(string `No employee information found for the user: ${userInfo.email}`);
```

apps/people-app/backend/modules/database/db_queries.bal
shayanmalinda: Having LIMIT 1 make the query more readable IMO. Anyway removed it.

Rashmika998 (approved):
LGTM

PR #51, opened 2025-10-20
[People App] Implement endpoints to fetch BUs, Teams, Sub Teams and Units

shayanmalinda:
Implement following new endpoints
- **GET** `business-units`
- **GET** `teams`
- **GET** `sub-teams`
- **GET** `units`

### DIFF
--- a/apps/people-app/backend/modules/database/types.bal
+++ b/apps/people-app/backend/modules/database/types.bal
@@ -178,7 +178,7 @@ public type SubTeam record {|
     string name;
 |};
 
-# Unit record type
+# Unit.
 public type Unit record {|
     # Unit ID
     int id;


apps/people-app/backend/modules/database/db_functions.bal
chanukaranaba: Without saying a generic error like **NoRowsError** can't we do, something like follow?
Also. No Rows means, it's not an error right?

` return error(EMPLOYEE_NOT_FOUND);`

chanukaranaba (approved):

apps/people-app/backend/modules/database/db_functions.bal
shayanmalinda: > Without saying a generic error like NoRowsError can't we do, something like follow? 
> Also. No Rows means, it's not an error right?

It’s not an `error`. It's handled as a `warning` and `http:NotFound` has been returned to the consumer.

`sql:NoRowsError` is used to do the type check and **it doesn't return anything**. Here, I’m returning `nil` instead of an `error`, so the parent function [1] can recognize the “_not found_” cases and return `http:NotFound` to the consumer. There I have logged a warning saying that "_Employee basic information not found_".

If we return something like `error(EMPLOYEE_NOT_FOUND)`, the parent function **won’t be able to distinguish** between real `errors` and “_not found_” cases, since both would be returned as `errors`.

Will proceed with merging the PR. Let me know if you have additional concerns.

[1] https://github.com/shayanmalinda/people-ops-suite/blob/6e7a9b8e02d1df32072ef1fe2e82d8332bc657b7/apps/people-app/backend/service.bal#L84

sacheeramesh:
### Comment

I don’t see a usage of the following endpoints within the **People App** context:

- `GET /business-units`
- `GET /teams`
- `GET /sub-teams`
- `GET /units`

Could you please explain the purpose of these?

Normally, the organization structure is a **predefined hierarchy**. Even if we store `Business Units`, `Teams`, `Sub Teams`, and `Units` in separate tables, their combinations are **static and predefined**.

Each **Business Unit** should have an `id`, `name`, and a predefined list of **Teams**.  
Each **Business Unit–Team** combination should have a predefined list of **Sub Teams**.  
Finally, each **Business Unit–Team–Sub Team** combination may (or may not) have an associated list of **Units**.

All these relationships are defined upfront, and typically don’t require dynamic endpoints for retrieval.

#### Example:

```ballerina
BusinessUnit[] businessUnits = [
    {
        id: 1,
        name: "Digital Transformation",
        teams: [
            {
                id: 101,
                name: "Internal Applications",
                subTeams: [
                    {
                        id: 1001,
                        name: "Employee Experience",
                        units: [
                            { id: 10001, name: "HR Tools" },
                            { id: 10002, name: "Onboarding Portal" }
                        ]
                    },
                    {
                        id: 1002,
                        name: "Automation & Integrations",
                        units: [
                            { id: 10003, name: "Workflow Automation" },
                            { id: 10004, name: "System Connectors" }
                        ]
                    }
                ]
            },
            {
                id: 102,
                name: "Data & Analytics",
                subTeams: [
                    {
                        id: 1003,
                        name: "Data Engineering",
                        units: [
                            { id: 10005, name: "ETL Pipelines" },
                            { id: 10006, name: "Data Quality" }
                        ]
                    },
                    {
                        id: 1004,
                        name: "Data Science",
                        units: [
                            { id: 10007, name: "Predictive Analytics" },
                            { id: 10008, name: "Visualization" }
                        ]
                    }
                ]
            }
        ]
    }
];

PR #253, opened 2026-04-07
[People App] feat: Implement bulk employee onboarding with CSV upload and validation

NethuliDulansa:
## Goals

- Allow admins to upload a CSV file containing multiple employee records and have all of them validated and created in a single request
- Provide clear per-row validation feedback so users can fix errors and re-upload without guessing what went wrong
- Auto-generate employee IDs for all employment types using in-memory sequence caching to minimize DB calls during batch processing
- Auto-assign houses based on least active employee count, removing the need for manual house selection during bulk onboarding

- Added POST /employees/bulk endpoint accepting multipart/form-data with a "file" field containing the CSV
- CSV is parsed using ballerina/data.csv directly into a typed BulkEmployeeCsvRow record, eliminating manual header mapping
- Processing runs in two passes — first pass validates all rows and detects intra-CSV duplicate emails and NIC/Passport numbers, second pass builds payloads and auto-generates employee IDs; all employees are inserted in a single DB transaction only if all rows pass validation
- Employee ID generation uses an in-memory context and sequence cache keyed.
- Replaced deprecated ballerina/regex with regexp:RegExp and .isFullMatch() throughout

> Frontend

- Added BulkOnboarding page with a two-panel layout — instructions with a downloadable CSV template on the left, drag-and-drop file upload on the right
- Client-side CSV parsing gives instant row count feedback before upload
- Validation errors returned from the backend are displayed in a grouped accordion UI, one card per row, showing the employee name alongside the row number and all field-level errors with the offending value included in the message
- Added Redux slice (bulkOnboardingSlice) handling loading, success, and per-row error states

image

image

image

## Summary by CodeRabbit

* **New Features**
  * Admin-only bulk employee CSV onboarding API with server-side validation, duplicate detection, balanced house assignment, and create/skip/error report.

* **Frontend**
  * Bulk Onboarding page: template download, drag-and-drop/file picker, CSV parsing, per-row expandable error details, success summary; onboarding split into Single + Bulk tabs and promoted to top-level route with updated tab styling.

* **Utilities**
  * CSV helpers, template constants, BOM handling, escaping, and row-counting utilities.


apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -529,6 +529,48 @@ isolated function getSubTeamsQuery(int? teamId = ()) returns sql:ParameterizedQu
...
+        } else {
+            inClause = sql:queryConcat(inClause, `, `, `${nics[i]}`);
+        }
+    }
+
+    return sql:queryConcat(
+            `SELECT nic_or_passport FROM personal_info WHERE nic_or_passport IN (`,
+            inClause,
+            `);`
+    );
```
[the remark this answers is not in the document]
NethuliDulansa: @coderabbitai db_functions.bal lines 159–161 and 172–174 already have if emails.length() == 0 { return []; } guards on the public functions. The query builders are private and unreachable with empty input. What do you think ?

apps/people-app/backend/service.bal
```diff
@@ -892,6 +891,180 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+                rollbackEmployeeCreation(emp.employeeId, emp.payload.workEmail);
+                provisioningErrors.push({
+                    employeeId: emp.employeeId,
+                    workEmail: emp.payload.workEmail,
+                    reason: scimResult.message()
+                });
+                continue;
+            }
+
+            string[]|error groupNames = database:getAsgardeoGroupsByEmploymentType(emp.payload.employmentTypeId);
```
shayanmalinda, line 1104: **⚠️ Consistency Gap: Asgardeo user is orphaned when group fetch (or group assignment) fails**

The flow here is: `scim:createUser` succeeds → `getAsgardeoGroupsByEmploymentType` (or one of the `addUserToGroup` calls below) fails → `rollbackEmployeeCreation` deletes the DB row, but the Asgardeo user lives on. The log message correctly says "requires manual cleanup", which is honest but means HR has to chase orphaned identities by hand. For a bulk endpoint, a network blip mid-processing could leave dozens of these.

If `scim` has a `deleteUser` (most SCIM clients do), invoking it inside the same rollback path would close the loop. If not, at least surfacing the orphan in the response (e.g. an `orphanedScimUsers` list) lets the operator find and clean them without spelunking logs.

Suggested fix:
```ballerina
error? scimRollback = scim:deleteUser(emp.payload.workEmail);
if scimRollback is error {
    log:printError("SCIM rollback failed after partial provisioning",
        scimRollback, workEmail = emp.payload.workEmail);
}
rollbackEmployeeCreation(emp.employeeId, emp.payload.workEmail);
```

---

apps/people-app/backend/service.bal
```diff
@@ -892,6 +891,180 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+        if payloadResult is error {
+            log:printError("Error building bulk employee payloads", payloadResult);
+            return <http:InternalServerError>{
+                body: {
+                    message: ERROR_EMPLOYEE_CREATION_FAILED
+                }
+            };
+        }
+
+        int|error created = database:addEmployeesBulk(
```
shayanmalinda: **⚠️ Race Condition: two concurrent bulk uploads can generate colliding employee IDs**

`buildBulkPayloads` calls `generateBulkEmployeeId`, which calls `database:getLastEmployeeNumericSuffix` *outside* any transaction. Two admins clicking Upload at the same time would both read the same `lastNumericId`, both build IDs `WSO2-101..N` in their in-memory caches, and both head into `addEmployeesBulk` with overlapping IDs. Whoever commits second hits a unique-constraint violation, the `retry transaction` re-runs with the same in-memory IDs (the cache isn't recomputed on retry), and the whole upload fails for that admin.

It's not a correctness bug (no data is corrupted thanks to the unique constraint), but it is a rough UX: the second admin has to retry manually with no clear reason in the response. Two options:

1. Move ID generation *inside* the `addEmployeesBulk` transaction so `SELECT ... FOR UPDATE` actually serializes concurrent callers, OR
2. On `retry transaction` invocation, re-run `buildBulkPayloads` (or at least clear and refill the sequence cache) so the second attempt picks up the freshly-incremented max.

Worth at minimum documenting the failure mode in the response (`return error("Concurrent upload conflict — please retry")`) when the underlying insert error is a unique-key collision.

---

apps/people-app/webapp/src/view/employees/onboarding/bulkOnboarding/BulkOnboarding.tsx
```diff
@@ -0,0 +1,1393 @@
...
+        setRowNames(buildRowNames(parsed));
+        setSelectedFile(file);
+      };
+      reader.onerror = () => {
+        setFileError("Failed to read the CSV file.");
+        setSelectedFile(null);
+        setRowCount(null);
+        setRowNames({});
+      };
+      reader.readAsText(file);
```
shayanmalinda, line 399: **⚠️ No file size / row limit on either client or server**

`FileReader.readAsText` loads the entire file into a browser string (UTF-16, so ~2 bytes per CSV character), and the backend's `parseCsvBytes` parses everything into a `BulkEmployeeCsvRow[]` in memory in one shot. A well-meaning admin pasting a 50 MB CSV will lock up Chrome; a malicious one can push the backend OOM with a much smaller multipart upload because of the doubled in-memory representation after parsing.

Cheap and effective fix on both ends:
```ts
// frontend, in processFile:
const MAX_BYTES = 5 * 1024 * 1024;
if (file.size > MAX_BYTES) {
  setFileError(`File too large (max ${MAX_BYTES / 1024 / 1024} MB)`);
  return;
}
```
```ballerina
// backend, after extractCsvFileBytes:
final int MAX_BULK_ROWS = 1000;
if rows.length() > MAX_BULK_ROWS {
    return <http:BadRequest>{ body: { message: string `CSV exceeds ${MAX_BULK_ROWS}-row limit` }};
}
```

---

apps/people-app/backend/utils.bal
```diff
@@ -151,3 +144,641 @@ isolated function rollbackEmployeeCreation(string employeeId, string workEmail)
...
+                    row: rowNum,
+                    'field: CSV_FIELD_WORK_EMAIL,
+                    message: string `Work email already exists: ${existing.trim()}`
+                });
+            }
+        }
+    }
+
+    if candidateNics.length() > 0 {
+        string[] existingNics = check database:getExistingNicOrPassport(candidateNics);
```
shayanmalinda, line 662: **⚠️ Inconsistent case handling: NIC and EPF dedup is case-sensitive, but work email isn't**

For work email the code does `normEmail = row.workEmail.trim().toLowerAscii()` for the in-CSV map, and the SQL query wraps both sides in `LOWER(...)`. For NIC and EPF the value is stored and compared as-is. The result: a row with `nicOrPassport="ABC123"` and a DB row with `"abc123"` will *not* be detected as a duplicate here, and the insert will then succeed or fail depending on the DB column collation. That divergence between layers will eventually bite.

Pick one rule and apply it everywhere. If NICs are conventionally uppercase, normalize to upper at both write and compare time; if the column collation is already case-insensitive (e.g. `utf8mb4_general_ci`), the in-memory map should be normalized to match.

Suggested fix (mirroring the email approach):
```ballerina
string normNic = row.nicOrPassport.trim().toLowerAscii();
// …
// db_queries.bal:
`SELECT LOWER(nic_or_passport) AS nic_or_passport FROM personal_info WHERE LOWER(nic_or_passport) IN (...)`
```

---

apps/people-app/backend/utils.bal
```diff
@@ -151,3 +144,641 @@ isolated function rollbackEmployeeCreation(string employeeId, string workEmail)
...
+isolated function loadBulkReferenceData() returns BulkRefData|error {
+    database:BusinessUnit[] businessUnits = check database:getBusinessUnits();
+    database:Team[] teams = check database:getTeams();
+    database:SubTeam[] subTeams = check database:getSubTeams();
+    database:Unit[] units = check database:getUnits();
+    database:Designation[] designations = check database:getDesignations();
+    database:EmploymentType[] employmentTypes = check database:getEmploymentTypes();
+    database:CompanyResponse[] companies = check database:getCompanies();
+    database:Office[] offices = check database:getOffices();
+    database:House[] houses = check database:getHouses();
```
shayanmalinda: **🧹 Performance: 9 serial DB round-trips for reference data on every bulk upload**

Each `check database:get…()` runs one after the next, even though they're independent. Wrapping them in `start` / `wait` (or the `wait {a, b, c, …}` form) drops the wall-clock cost to roughly the slowest single query — useful both for the bulk endpoint and for any future caller of `loadBulkReferenceData`.

Suggested fix:
```ballerina
future<database:BusinessUnit[]|error> bu = start database:getBusinessUnits();
future<database:Team[]|error> t  = start database:getTeams();
// … one per source
database:BusinessUnit[] businessUnits = check wait bu;
database:Team[] teams = check wait t;
// …
```

> 📖 **Reference ([using_query_expressions_for_iterations](https://learn-ballerina.github.io/best_practices/using_query_expressions_for_iterations.html)):**
> "Replace `foreach` loops with query expressions when filtering, transforming, or sorting"

(Also: `buildBulkPayloads` re-fetches houses via `getHousesWithActiveEmployeeCounts` even though `loadBulkReferenceData` already loaded houses for ID lookup — collapsing both into the count-bearing call would save one more round-trip.)

---

apps/people-app/backend/modules/database/db_functions.bal
```diff
@@ -352,6 +401,32 @@ public isolated function deleteEmployeeById(string employeeId) returns error? {
...
+
+    retry transaction {
+        foreach int i in 0 ..< payloads.length() {
+            CreateEmployeePayload payload = payloads[i];
+            string employeeId = employeeIds[i];
+            int personalInfoId = check addPersonalInfo(payload.personalInfo, createdBy);
+            _ = check addEmployeeRecord(payload, createdBy, personalInfoId, employeeId);
+            check syncEmergencyContacts(employeeId, payload.personalInfo.emergencyContacts ?: [], createdBy);
+            check syncAdditionalManagers(employeeId, payload.additionalManagerEmails, createdBy);
+        }
```
shayanmalinda, line 463: **🧹 Performance: per-row inserts inside the transaction — N×4 statements**

For every employee the loop fires four sequential SQL statements (`addPersonalInfo`, `addEmployeeRecord`, `syncEmergencyContacts`, `syncAdditionalManagers`). At 100 rows that's 400 round-trips inside a single transaction — workable, but the transaction holds locks for noticeably longer than it needs to.

If the `addPersonalInfo` and `addEmployeeRecord` queries can be reworked into multi-row `INSERT ... VALUES (...), (...), ...` plus a single `INSERT` for emergency contacts and additional managers, both elapsed time and lock-contention with single-employee creates would drop meaningfully. Not a blocker for the first cut — flagging for follow-up if bulk-onboarding sizes grow.

---

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -561,6 +561,68 @@ isolated function getSubTeamsQuery(int? teamId = ()) returns sql:ParameterizedQu
     return sql:queryConcat(query, `;`);
 }
 
+# Fetch existing work emails.
+#
+# + emails - Work email list
+# + return - Query to fetch existing work emails
+isolated function getExistingWorkEmailsQuery(string[] emails) returns sql:ParameterizedQuery {
```
shayanmalinda, line 595: **🧹 Nit: IN-clause builder duplicates the same imperative pattern across three queries**

`getExistingWorkEmailsQuery`, `getExistingNicOrPassportQuery`, and `getExistingEpfsQuery` each repeat the same `foreach` + `if i == 0` ceremony to build the parameterised IN-list. (Parameterisation itself is fine — values are still bound, not concatenated.) Factoring this into a tiny helper keeps the three queries readable and avoids drifting if a fourth gets added later.

Suggested helper:
```ballerina
isolated function buildInClause(string[] values) returns sql:ParameterizedQuery {
    sql:ParameterizedQuery clause = ``;
    foreach int i in 0 ..< values.length() {
        clause = i == 0
            ? sql:queryConcat(clause, `${values[i]}`)
            : sql:queryConcat(clause, `, `, `${values[i]}`);
    }
    return clause;
}
```

> 📖 **Reference ([using_query_expressions_for_iterations](https://learn-ballerina.github.io/best_practices/using_query_expressions_for_iterations.html)):**
> "Replace `foreach` loops with query expressions when filtering, transforming, or sorting"

---

apps/people-app/backend/constants.bal
```diff
@@ -19,3 +19,43 @@ const ERROR_PERSONAL_INFO_UPDATE_FAILED = "Failed to update personal information
 const ERROR_EMPLOYEE_INFO_UPDATE_FAILED = "Failed to update employee job information";
 const ERROR_EMPLOYEE_CREATION_FAILED = "Error occurred while creating a new employee";
 const WARNING_GROUP_ASSIGNMENT_FAILED = "Employee created successfully, but some group assignments failed";
+const MAX_BULK_ROWS = 1000;
```
shayanmalinda: We don't need this much, let's keep it as 100.
```suggestion
const MAX_BULK_ROWS = 100;
```

shayanmalinda: BTW, can you also work on testing a 100 BULK onboard? And see if there will be any gateway timeouts.

apps/people-app/backend/service.bal
```diff
@@ -892,6 +891,200 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+            return <http:Forbidden>{
+                body: {
+                    message: "You are not authorized to bulk create employees"
+                }
+            };
+        }
+
+        byte[]|http:BadRequest fileBytes = extractCsvFileBytes(req);
+        if fileBytes is http:BadRequest {
+            return fileBytes;
```
shayanmalinda, line 1003: Let's have an error log.

apps/people-app/backend/service.bal
```diff
@@ -892,6 +891,200 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+            log:printError("Error parsing bulk upload CSV", rows);
+            return <http:BadRequest>{
+                body: {
+                    message: "Invalid CSV file"
+                }
+            };
+        }
+
+        if rows.length() > MAX_BULK_ROWS {
+            return <http:BadRequest>{ body: { message: string `CSV exceeds ${MAX_BULK_ROWS}-row limit` }};
```
shayanmalinda: Let's have a warn log here.

apps/people-app/backend/service.bal
```diff
@@ -892,6 +891,200 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+        BulkPayloadResult payloadResult = buildBulkPayloads(firstPass.rowInfos, refData);
+
+        string[]|error generatedIds = database:addEmployeesBulk(
+                payloadResult.employees.map(e => e.payload),
+                userInfo.email);
+        
+        if generatedIds is error {
+            log:printError("Error occurred while bulk creating employees", generatedIds);
+            
+            if generatedIds.message().includes("Duplicate entry") || generatedIds.message().includes("Unique constraint") {
```
shayanmalinda: Capturing errors via substring matching isn't a reliable approach. If the Ballerina module changes and this message gets updated, the matching will silently break.

Instead, if possible, let's have the `database:addEmployeesBulk` function return a `map<int>` containing the `generatedId` and an `errorCode` for each entry. For example:

```json5
{
    "LK10001": 1000,    // SUCCESS
    "LK10002": 10001,   // DUPLICATE
    "LK10003": 10002,   // UNIQUE CONSTRAINT
    ...
}
```

Feel free to decide on the error codes as you see fit.

apps/people-app/backend/service.bal
```diff
@@ -892,6 +891,200 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+        database:BulkGroupAssignmentWarning[] groupAssignmentWarnings = [];
+        database:OrphanedScimUser[] orphanedScimUsers = [];
+
+        foreach ResolvedEmployee emp in payloadResult.employees {
+            error? scimResult = scim:createUser({
+                userName: string `${scim:asgardeoUserStoreDomain}/${emp.payload.workEmail}`,
+                emails: [emp.payload.workEmail],
+                name: {givenName: emp.payload.firstName, familyName: emp.payload.lastName},
+                urn\:scim\:wso2\:schema: {askPassword: true}
+            });
```
shayanmalinda: This is a SCIM-related implementation, so `service.bal` shouldn't hold this logic. Let's move it into the `scim` module.

The call site should simply invoke a function like the one below, and the rest should be handled inside the `scim` module:

```ballerina
error? scimResult = scim:createUser({
    userEmail: emp.payload.workEmail,
    firstName: emp.payload.firstName,
    lastName: emp.payload.lastName
});
```

or
```ballerina
error? scimResult = scim:createUser(emp.payload);
```

`service.bal` should only contain **business logic**.

apps/people-app/backend/types.bal
```diff
@@ -83,3 +83,151 @@ type EmployeeCreatedResponse record {|
...
+|};
+
+# Result of the second pass containing all resolved employees ready for bulk DB insertion.
+type BulkPayloadResult record {|
+    # Resolved employees with generated IDs and complete payloads
+    ResolvedEmployee[] employees;
+|};
+
+# Typed record mapped directly from a bulk employee onboarding CSV row.
+type BulkEmployeeCsvRow record {|
```
shayanmalinda, line 158: Why are we introducing a new Record type for this? Can't we re-use the already existing Single onboard records?

apps/people-app/webapp/src/config/constant.ts
```diff
@@ -302,3 +302,111 @@ export const Countries = [
...
+  "",
+  "",
+  "",
+  "",
+  "",
+  "",
+  "",
+  "",
+  "",
+  "manager1@wso2.com;manager2@wso2.com",
```
shayanmalinda: Won't it be better to use `,` instead? Hope these are additional managers?
```suggestion
  "manager1@wso2.com,manager2@wso2.com",
```

apps/people-app/backend/types.bal
NethuliDulansa: `BulkEmployeeCsvRow` cannot reuse `CreateEmployeePayload` because CSV parsing produces all-string data, whereas `CreateEmployeePayload` has typed fields (int IDs, Email[], EmergencyContact[], etc.).

The two-pass design separates:
 (1) raw CSV strings → validation/lookup, then (2) conversion → typed `CreateEmployeePayload`. 
 This ensures proper format conversion and DB consistency before insertion.

shayanmalinda (approved):

PR #263, opened 2026-04-17
[People App] feat: Add resignation and employment dates to employee records

NethuliDulansa:
## Purpose

This pull request adds support for managing and displaying employee leaver information, specifically resignation and final employment dates, across the database, backend, and frontend of the people app. The changes introduce new fields for resignation date, final working date, and final employment date, and ensure these are handled consistently.

## Goals

**Database schema and backend API enhancements:**

* Added `resignation_date`, `final_working_date`, and `final_employment_date` columns to the `employee` table, with corresponding migration scripts and updates to the employee creation SQL. 

**Frontend and form updates:**

* Extended TypeScript types and Redux slice (`Employee`, `UpdateEmployeeJobInfoPayload`, `CreateEmployeeFormValues`) to include the new leaver fields.

**UI enhancements:**

* Added a "Leaver Information" section to the job info step in the onboarding/edit employee flow, with date pickers for resignation date, final working date, and final employment date, including appropriate icons and helper text. 

## Summary by CodeRabbit

* **New Features**
  * Introduced employee status field with three states: Active, Left, and Marked leaver
  * Added resignation management with final working date, employment end date, and reason for leaving
  * Resignation details are only editable when employee status is marked as Left
  * Enhanced employee records with comprehensive leaver information and validation


shayanmalinda (commented):
Overall direction is good — consolidating departure data into the `employee` table simplifies queries. A few issues to address before merging.

apps/people-app/backend/resources/people_app_table_update_v1.0.14.sql
```diff
@@ -0,0 +1,145 @@
+ALTER TABLE `employee`
+    ADD COLUMN `last_working_date`    DATE         NULL AFTER `agreement_end_date`,
+    ADD COLUMN `employment_end_date`  DATE         NULL AFTER `last_working_date`,
+    ADD COLUMN `reason_for_leaving`   VARCHAR(300) NULL AFTER `employment_end_date`;
```
shayanmalinda: **🚨 Critical: Missing Data Migration**

This migration adds new columns to the `employee` table but doesn't copy existing data from the `resignation` table across. Any employee with resignation data already recorded will have it silently lost when this runs on production — and there's no way to recover it after the fact.

Suggested fix: Add this before the trigger changes to backfill the existing data:
```sql
UPDATE employee e
JOIN resignation r ON r.employee_id = e.id
SET
  e.last_working_date   = r.final_day_in_office,
  e.employment_end_date = r.final_day_of_employment,
  e.reason_for_leaving  = r.reason;
```

---

apps/people-app/backend/resources/people_app_table_update_v1.0.14.sql
```diff
@@ -0,0 +1,145 @@
...
+    NEW.last_working_date,         NEW.employment_end_date,  NEW.reason_for_leaving,
+    NEW.employment_type_id,        NEW.designation_id,       NEW.company_id,
+    NEW.office_id,                 NEW.team_id,              NEW.sub_team_id,
+    NEW.business_unit_id,          NEW.unit_id,              NEW.house_id,
+    NEW.personal_info_id,
+    NEW.created_by,                NEW.created_on,
+    NEW.updated_by,                NEW.updated_on
+  );
+END//
+DELIMITER ;
```
shayanmalinda: **🚨 Critical: Orphaned `resignation` Table**

The migration removes all references to the `resignation` table from queries, but never drops the table itself. This leaves a stale table in the schema that nothing reads from or writes to — which can confuse future developers and make it unclear whether it's safe to remove.

Suggested fix: Add this at the end of the migration (after the data has been copied across):
```sql
DROP TABLE IF EXISTS resignation;
```

---

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -1215,6 +1216,40 @@ isolated function updateEmployeeJobInfoQuery(string employeeId, UpdateEmployeeJo
...
+            updates.push(`reason_for_leaving = ${payload.reasonForLeaving}`);
+        }
+    }
+
+    if payload.employeeStatus is EmployeeStatus {
+        updates.push(`employee_status = ${payload.employeeStatus}`);
+    } else {
+        boolean hasLeaverDate = (payload.lastWorkingDate is string && payload.lastWorkingDate != "") ||
+                                (payload.employmentEndDate is string && payload.employmentEndDate != "");
+        if hasLeaverDate {
```
shayanmalinda: **⚠️ Major: Employee Status Can't Be Reverted to Active**

When an admin clears both date fields (setting them back to `NULL`), `hasLeaverDate` becomes `false` and no status update is pushed — so the employee stays permanently marked as `Left` with no way to recover through this flow. This could cause issues in reports, permissions, and any system that relies on `employee_status`.

Suggested fix: Add an `else` branch to reset the status when the dates are cleared:
```ballerina
} else {
    updates.push(`employee_status = ${EMPLOYEE_ACTIVE}`);
}
```

---

apps/people-app/backend/modules/database/utils.bal
```diff
@@ -369,10 +369,9 @@ public isolated function buildResignationCsv(Employee[] employees, map<string> n
             csvEscape(e.office),
```
shayanmalinda: **💡 Nitpick: Misleading Function Name**

Minor point — `buildResignationCsv` now reads from `employee` fields rather than the `resignation` table, so the name no longer reflects what the function does. Someone reading it in the future may assume it's tied to the old resignation flow.

Suggested fix: Rename to `buildLeaverCsv`.

---

apps/people-app/webapp/src/slices/employeeSlice/employee.ts
```diff
@@ -26,6 +26,8 @@ import { enqueueSnackbarMessage } from "@slices/commonSlice/common";
 import { APIService } from "@utils/apiService";
 import { HttpStatusCode, isCancel } from "axios";
```
shayanmalinda, line 28: **💡 Nitpick: Redundant Re-export**

Minor point — `EmployeeStatus` is imported on line 17 for local use and also re-exported here. Without a comment, it's not clear whether this is intentional (to maintain backward compatibility for existing consumers of this module) or accidental. A future developer might remove it thinking it's a mistake.

Suggested fix: If intentional, add a short comment explaining why:
```ts
// Re-exported for backward compatibility with existing consumers of this module
export { EmployeeStatus } from "@/types/types";
```

---

apps/people-app/backend/resources/people_app_table_update_v1.0.14.sql
shayanmalinda: @NethuliDulansa just realised that we don't need to introduce new columns to capture these informatoin. It seems like we need to consume the resignation table for this purpose.

apps/people-app/backend/resources/people_app_table_update_v1.0.14.sql
shayanmalinda: We'll have to keep the resignation table as it is.

apps/people-app/webapp/src/view/employees/employeeOnboarding/steps/JobInfo.tsx
```diff
@@ -1364,8 +1367,28 @@ export default function JobInfoStep({ isEditMode }: { isEditMode?: boolean }) {
...
+                  </MenuItem>
+                ))}
+              </TextField>
+            </Grid>
+          ) : null}
           {isInternship ? (
-            <Grid item xs={12} sm={6} md={4}>
+            <Grid item xs={12} sm={6} md={3}>
               <TextField
                 select
```
shayanmalinda, line 1413: **⚠️ Major: Stale Leaver Dates Persist When Status Is Changed Back**

The leaver fields are disabled in the UI when status isn't `Left`, but their values aren't cleared when the status changes. If someone selects `Left`, fills in the dates, then switches back to `Active`, those dates are still in the form and will be saved to the database on submit — leaving an Active employee with leaver dates set against them.

Suggested fix: Clear the leaver fields when status changes away from `Left`:
```tsx
onChange={(e) => {
  setFieldValue("employeeStatus", e.target.value);
  if (e.target.value !== EmployeeStatus.Left) {
    setFieldValue("lastWorkingDate", null);
    setFieldValue("employmentEndDate", null);
    setFieldValue("reasonForLeaving", null);
  }
}}
```

---

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -1215,6 +1216,40 @@ isolated function updateEmployeeJobInfoQuery(string employeeId, UpdateEmployeeJo
         }
     }
 
+    if payload.lastWorkingDate is string {
+        if payload.lastWorkingDate == "" {
```
shayanmalinda: **💡 Nitpick: Empty-String Check Is Unreachable**

Minor point — the `== ""` branch for the date fields can never actually run. The `@constraint:String {pattern: ...}` annotation on `UpdateEmployeeJobInfoPayload` rejects empty strings at the validation layer before this code is reached, so this branch is effectively dead code.

Suggested fix: Remove the empty-string branches to simplify the logic. If they're intentionally kept as a safety net for internal calls that bypass validation, add a comment explaining that.

---

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -1215,6 +1216,40 @@ isolated function updateEmployeeJobInfoQuery(string employeeId, UpdateEmployeeJo
...
+    if payload.reasonForLeaving is string {
+        if payload.reasonForLeaving == "" {
+            updates.push(`reason_for_leaving = NULL`);
+        } else {
+            updates.push(`reason_for_leaving = ${payload.reasonForLeaving}`);
+        }
+    }
+
+    if payload.employeeStatus is EmployeeStatus {
+        updates.push(`employee_status = ${payload.employeeStatus}`);
```
shayanmalinda, line 1224: **⚠️ Major: Setting Dates Without a Status Silently Changes the Employee's Status**

When `employeeStatus` isn't provided in the update payload but leaver dates are, the backend automatically sets `employee_status = 'Left'` without the caller knowing. This hidden side-effect can cause unexpected status changes — for example, if someone pre-fills dates as a draft before officially marking an employee as a leaver.

Suggested fix: Either require the status to always be set explicitly, or move this logic into a clearly named helper (e.g. `deriveEmployeeStatus`) and document the behaviour on the payload type so callers know to expect it.

---

NethuliDulansa: The concern is not present in the current code. No fix is needed.

apps/people-app/backend/modules/database/db_functions.bal
```diff
@@ -524,10 +524,42 @@ public isolated function updateEmployeeJobInfo(string employeeId, UpdateEmployee
...
+# + payload - Job information update payload
+# + updatedBy - User performing the operation
+# + return - Nil or error
+isolated function syncResignationRecord(string employeeId, UpdateEmployeeJobInfoPayload payload, string updatedBy)
+    returns error? {
+
+    EmployeeStatus? newStatus = payload.employeeStatus;
+
+    if newStatus is EmployeeStatus && newStatus != EMPLOYEE_LEFT {
+        _ = check databaseClient->execute(deleteResignationQuery(employeeId));
```
shayanmalinda: I assume the current behavior is that when a resigned employee is reactivated, the existing resignation record is deleted. If that is the case, I don’t think we should remove that information. It would be better to retain the resignation details for future reference.

shayanmalinda: If the employee resigns again, we can upsert the resignation details.

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -1226,6 +1235,41 @@ isolated function updateEmployeeJobInfoQuery(string employeeId, UpdateEmployeeJo
...
+# + updatedBy - User performing the update
+# + return - sql:ParameterizedQuery - Upsert query for the resignation table
+isolated function upsertResignationQuery(string employeeId, UpdateEmployeeJobInfoPayload payload, string updatedBy)
+    returns sql:ParameterizedQuery {
+
+    string? finalDayInOffice = payload.finalDayInOffice;
+    string? finalDayOfEmployment = payload.finalDayOfEmployment;
+    string? resignationReason = payload.resignationReason;
+
+    return `INSERT INTO resignation (employee_id, final_day_in_office, final_day_of_employment, reason, created_by, updated_by)
```
shayanmalinda: **⚠️ Major: `resignation.date` column is not set in this INSERT — `resignationDate` may always be NULL**

The INSERT column list is `(employee_id, final_day_in_office, final_day_of_employment, reason, created_by, updated_by)` — the `date` column is absent. The `ON DUPLICATE KEY UPDATE` clause also doesn't touch it. Both `getEmployeeInfoQuery` and `getEmployeesQuery` fetch `r.date AS resignationDate` and return it to clients, so if `date` has no `DEFAULT` in the schema, every record created by this upsert will have `resignationDate = NULL`.

Question: Does `resignation.date` have a `DEFAULT CURRENT_TIMESTAMP` defined in the migration script (which isn't in this diff)? If not, this is a silent data gap — the field will appear in API responses but never be populated.

Suggested fix: If `date` is meant to record when the resignation was entered, add it to the INSERT and the UPDATE clause:
```sql
INSERT INTO resignation (employee_id, date, final_day_in_office, ...)
    SELECT e.id, CURRENT_TIMESTAMP(6), ...
    ON DUPLICATE KEY UPDATE
        date = IF(date IS NULL, CURRENT_TIMESTAMP(6), date), ...
```

---

apps/people-app/webapp/src/view/employees/employeeOnboarding/steps/JobInfo.tsx
```diff
@@ -1668,6 +1699,87 @@ export default function JobInfoStep({ isEditMode }: { isEditMode?: boolean }) {
           </Grid>
         </Grid>
       </Box>
+
+      {isEditMode ? (
+        <Box>
+          <SectionHeader
+            icon={SECTION_ICONS.leaver}
+            title="Resignation Details"
```
shayanmalinda, line 1719: **❓ Needs Clarification: Are the new resignation fields covered by the Yup validation schema?**

This section adds `finalDayInOffice`, `finalDayOfEmployment`, and `resignationReason` to the Formik form, but the diff doesn't include an update to the Yup validation schema. Without schema entries, these fields won't show inline validation errors if a user submits them in an invalid state — for example, a non-date string in a date field, or a reason that's too long.

Question: Is the Yup schema updated elsewhere (e.g. in a separate validation file not in this diff), or are these fields intentionally left unvalidated at the form level and relying solely on the backend to reject bad input?

Suggested fix: If they're not in the schema yet, add entries — at minimum a date format check for the two date pickers and a `max()` length check for the reason field to match the `maxLength={300}` on the text input.

---

apps/people-app/backend/modules/database/types.bal
```diff
@@ -902,6 +902,15 @@ public type UpdateEmployeeJobInfoPayload record {|
...
     # Employee Status
     EmployeeStatus? employeeStatus = ();
+    # Final day in office
+    @constraint:String {pattern: re `${DATE_PATTERN_STRING}`}
+    string? finalDayInOffice = ();
+    # Final day of employment 
+    @constraint:String {pattern: re `${DATE_PATTERN_STRING}`}
+    string? finalDayOfEmployment = ();
+    # Resignation reason
+    @constraint:String {pattern: re `${DATE_PATTERN_STRING}`}
```
shayanmalinda: We don't need a Date constraint for the resignation reason. Please fix.

shayanmalinda (approved):

PR #267, opened 2026-04-21
[People App] Implement SCIM operations for user provisioning and group management

NethuliDulansa:
This pull request introduces SCIM (System for Cross-domain Identity Management) user provisioning and group assignment to the employee creation flow in the People App backend. It adds a new mapping between employment types and Asgardeo (IDP) groups, provisions new employees in Asgardeo via the SCIM operations service, and assigns them to the correct groups based on their employment type. The changes include database schema updates, new backend modules, and integration into the employee creation API.

**SCIM Integration and Employee Provisioning:**

* The employee creation API (`service.bal`) now provisions a user in Asgardeo via the SCIM operations service before creating the employee record. If user provisioning fails, employee creation is aborted. After a successful employee record creation, the user is added to all Asgardeo groups mapped to their employment type. Errors are logged and do not block the main flow, but warnings are issued if group assignment fails.

* New backend module `scim_operations` is introduced, including:
  - [`client.bal`](diffhunk://#diff-ca28f8879445165eec73522a18d06589340d3d22aab2b0f9a3d027d9a651bdaaR1-R26): Configures and initializes the SCIM operations service HTTP client.
  - [`scim_operations.bal`](diffhunk://#diff-94487ad6cf1039eb694f14f95b5807286e6dbddf2d4189f2ded2d6ad2a30090cR1-R44): Implements `createUser` and `addUserToGroup` functions for user provisioning and group assignment.
  - [`types.bal`](diffhunk://#diff-1dd65182f7fed17b4bd09c141168d9f51e8ca338995c81704f3e6a083df4a6e1R1-R78): Defines types for SCIM user creation, bulk responses, and group assignment responses.

**Database and Schema Changes:**

* Adds a new mapping table `employment_type_idp_group` to associate employment types with Asgardeo group names. This is reflected in both the main schema and migration scripts.
* Drops the mapping table in the schema teardown script to support clean database resets.

**Employment Type to Group Mapping Logic:**

* Adds `getAsgardeoGroupsForEmploymentType` function in `db_functions.bal` and its corresponding SQL query in `db_queries.bal` to fetch Asgardeo group names for a given employment type. 

**Configuration:**

* Adds a new configuration section for the SCIM operations service URL in `Config.toml.local`.

**Module Integration:**

* Imports the new `scim_operations` module in the main service.

## Summary by CodeRabbit

* **New Features**
  * Employees are provisioned in the identity system on creation and auto-assigned to identity groups based on employment type.
  * Employee deletion performs hard deletes of related records and audit entries with retry/compensation behavior.

* **Chores**
  * Added configurable settings for SCIM/identity operations and an internal SCIM client.
  * Database migration: added employment_type-to-group mapping table.


apps/people-app/backend/service.bal
```diff
@@ -965,17 +966,54 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+                            workEmail = payload.workEmail, group = groupName);
+                    continue;
+                }
+                if addResult.failedUsers.length() > 0 {
+                    log:printWarn("Asgardeo reported failure adding user to group",
+                            workEmail = payload.workEmail, group = groupName,
+                            failedUsers = addResult.failedUsers);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
NethuliDulansa: @coderabbitai Addressed this comment. Please check and confirm.”

apps/people-app/backend/resources/people_app_creation.sql
```diff
@@ -552,6 +553,23 @@ CREATE TABLE `employee_additional_managers_audit` (
...
+  UNIQUE KEY `uniq_etag_employment_type_group` (`employment_type_id`, `group_name`),
+  KEY `idx_etag_employment_type_id` (`employment_type_id`),
+  CONSTRAINT `fk_etag_employment_type`
+    FOREIGN KEY (`employment_type_id`) REFERENCES `employment_type` (`id`)
+    ON DELETE CASCADE ON UPDATE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+
 -- Procedure: prc_personal_info_audit
 DELIMITER //
 CREATE PROCEDURE `prc_personal_info_audit`(
```
shayanmalinda, line 575: **⚠️ Issue: Collation mismatch with existing tables**

This new table uses `utf8mb4_unicode_ci`, but every other table in this creation script uses `utf8mb4_0900_ai_ci` (see the line just above: `COLLATE = utf8mb4_0900_ai_ci`). Mismatched collations can cause `Illegal mix of collations` errors when MySQL needs to compare or JOIN strings across these tables — for example, if a query ever joins `employment_type_idp_group.group_name` with a column from another table. The same mismatch is in the migration file `people_app_table_update_v1.0.14.sql`.

Suggested fix: Use the same collation as the rest of the schema:
```sql
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```

---

apps/people-app/backend/modules/scim_operations/client.bal
```diff
@@ -0,0 +1,26 @@
...
+import ballerina/http;
+
+configurable string scimOperationsServiceUrl = "";
+public configurable string asgardeoUserStoreDomain = "DEFAULT";
+
+@display {
+    label: "SCIM Operations Service Client",
+    id: "scim_operations/client"
+}
+final http:Client scimOperationsClient = check new (scimOperationsServiceUrl);
```
shayanmalinda: **🤔 Question: No auth configuration on the SCIM operations client**

Every other HTTP client in this project (QR portal, wallet, transaction, etc.) is created with `auth: {...clientAuthConfig}` for OAuth2, but this client has no auth at all. If the SCIM operations service is protected by OAuth2 or any other mechanism, requests will fail with 401. Is auth handled at the network/service-mesh level for this service, or does it need credentials added here?

Suggested fix: If auth is needed, add it to the config and client:
```ballerina
final http:Client scimOperationsClient = check new (scimOperationsServiceUrl, {
    auth: {...clientAuthConfig}
});
```

---

apps/people-app/backend/modules/database/db_functions.bal
```diff
@@ -321,6 +332,28 @@ public isolated function addEmployee(CreateEmployeePayload payload, string creat
     return lastInsertedId;
 }
 
+# Compensating delete used to roll back `addEmployee` when downstream provisioning fails.
+#
+# + employeeId - Employee ID string
+# + return - Nil on success or error
+public isolated function deleteEmployeeById(string employeeId) returns error? {
+    retry transaction {
+        int personalInfoId = check databaseClient->queryRow(
```
shayanmalinda: **📝 Nitpick: Two separate queries for the same row can be combined into one**

Minor one — these two `queryRow` calls both hit the same `employee` row by `employee_id`. Combining them into a single query saves a DB round-trip, which matters in a compensating delete where speed reduces the inconsistency window.

Suggested fix:
```ballerina
record {|int personalInfoId; int employeePkId;|} ids = check databaseClient->queryRow(
    `SELECT personal_info_id AS personalInfoId, id AS employeePkId
     FROM employee WHERE employee_id = ${employeeId}`);
```
Then use `ids.personalInfoId` and `ids.employeePkId` in the subsequent delete calls.

> 📖 **Reference ([Avoid redundant variables](https://learn-ballerina.github.io/best_practices/avoid_redundant_variables.html)):**
> "Eliminate variables used only once — return the value directly"

---

shayanmalinda: +1, but let's have a record type instead of defining attributes inline.

apps/people-app/backend/modules/scim/client.bal
```diff
@@ -0,0 +1,31 @@
...
+
+@display {
+    label: "SCIM Operations Service Client",
+    id: "scim_operations/client"
+}
+final http:Client scimOperationsClient = check new (scimOperationsServiceUrl, {
+    auth: {
+        ...scimClientAuthConfig
+    }
+});
```
shayanmalinda, line 43: **⚠️ Issue: SCIM client is missing retry config, timeout, and HTTP settings that other people-app clients use**

Comparing with the existing clients in this project: the QR client (`modules/qr/clients.bal`) uses `retryConfig` (3 attempts, 2s interval, targeting 502/503/504), `httpVersion: http:HTTP_1_1`, and `http1Settings: {keepAlive: http:KEEPALIVE_NEVER}`. The visitor-app email client adds a `timeout: 10.0`. The new SCIM client has none of these. Since SCIM user provisioning is a critical path (employee creation fails and rolls back if this call fails), transient 502/503/504 errors from the upstream service would cause unnecessary rollbacks without retries.

Suggested fix: Add retry and timeout config consistent with the QR client pattern:
```ballerina
final http:Client scimOperationsClient = check new (scimOperationsServiceUrl, {
    auth: {
        ...scimClientAuthConfig
    },
    timeout: 15.0,
    httpVersion: http:HTTP_1_1,
    http1Settings: {keepAlive: http:KEEPALIVE_NEVER},
    retryConfig: {
        count: 3,
        interval: 3.0,
        statusCodes: [
            http:STATUS_BAD_GATEWAY,
            http:STATUS_SERVICE_UNAVAILABLE,
            http:STATUS_GATEWAY_TIMEOUT
        ]
    }
});
```

---

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -1694,3 +1703,61 @@ isolated function getParkingReservationsByEmployeeQuery(string employeeEmail, st
 # + return - Parameterized query returning work_email and full_name columns
 isolated function getEmployeeEmailToNameMapQuery() returns sql:ParameterizedQuery =>
     `SELECT work_email, CONCAT(first_name, ' ', last_name) AS full_name FROM employee;`;
+
+# Delete additional managers for an employee.
+#
```
shayanmalinda, line 1708: **⚠️ Issue: Compensating delete uses hard `DELETE` but the codebase uses soft deletes for the same tables**

The existing codebase uses soft deletes for these same tables — `deleteEmergencyContactQuery` sets `is_active = 0`, and `deleteAdditionalManagerQuery` does the same. These soft deletes trigger UPDATE-based audit triggers (which detect `is_active` flipping from 1 to 0 and log a `'DELETE'` action type). The new compensating-delete queries here use hard `DELETE FROM`, which will bypass the audit triggers entirely. This means a rolled-back employee creation won't appear in the audit trail, making it invisible to compliance or debugging.

Suggested fix: Either (a) use the existing soft-delete pattern (`SET is_active = 0`) so the audit triggers fire and the rollback is traceable, or (b) if hard delete is intentional for compensating rollbacks (since the record never fully existed), add a comment explaining why audit trail is skipped.

---

apps/people-app/backend/service.bal
```diff
@@ -974,8 +975,49 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+                    groupNames, workEmail = payload.workEmail, employmentTypeId = payload.employmentTypeId);
+            return newEmployeeId;
+        }
+        foreach string groupName in groupNames {
+            scim_operations:AddUsersToGroupResponse|error addResult =
+                    scim_operations:addUserToGroup(groupName, payload.workEmail);
+            if addResult is error {
+                log:printError("Failed to add user to Asgardeo group", addResult,
+                        workEmail = payload.workEmail, group = groupName);
+                continue;
```
shayanmalinda: **🤔 Design question: Partial group failures are invisible to the API caller**

If an employee's employment type maps to 5 groups and 3 fail, the caller gets a 200 with `newEmployeeId` and no indication that group assignment was partial. The errors are logged server-side, but the API consumer (likely the admin UI) has no way to know something went wrong. This could leave new employees without the access they need, with no one aware of it until they try to log in.

Suggested fix: Consider returning a richer response that surfaces warnings:
```ballerina
return {
    employeeId: newEmployeeId,
    warnings: failedGroups  // string[] of group names that failed
};
```
This would require updating the return type from `int` to a record type.

---

apps/people-app/backend/modules/database/db_functions.bal
```diff
@@ -245,6 +245,17 @@ public isolated function getEmploymentTypes() returns EmploymentType[]|error {
         select employmentType;
 }
 
+# Fetch IDP group names mapped to an employment type.
+#
+# + employmentTypeId - Employment type ID
+# + return - Group names, or error
+public isolated function getAsgardeoGroupsForEmploymentType(int employmentTypeId) returns string[]|error {
```
Rashmika998: [Minor] Wouldn't it better to rename this as `getGroupsByEmploymentType`? If we need the Asgardeo part then it should be `getAsgardeoGroupsByEmploymentType`

apps/people-app/backend/modules/scim/client.bal
Rashmika998, line 1: [Minor] We could rename this module as `scim` instead of `scim_operations` which aligns with our other apps

apps/people-app/backend/modules/scim/types.bal
```diff
@@ -0,0 +1,88 @@
...
+// under the License.
+
+# OAuth2 client auth configurations.
+public type ClientAuthConfig record {|
+    # Token URL
+    string tokenUrl;
+    # Client Id
+    string clientId;
+    # Client Secret
+    string clientSecret;
```
Rashmika998, line 24: We need to add scopes(`string []`) in here as the endpoints are restricted by scopes

apps/people-app/backend/modules/database/db_queries.bal
NethuliDulansa: Intentionally keeping hard DELETE here — these queries are only used as compensating deletes when onboarding fails mid-way, so the record never fully existed. Using soft delete would pollute is_active = 0 queries with incomplete employee records and create misleading audit trail entries for employees that were never successfully provisioned. The rollback is traceable via the application-level error logs.

shayanmalinda (approved):

PR #280, opened 2026-05-18
[Meals App] (webapp) fix: Refactor Micro App Logic and Enhance Auth Handling

dinethde:
This pull request refactors how the application determines and uses micro app mode. The main change is the removal of the `useMicroApp` hook in favor of directly using the `isMicroApp` boolean from the config throughout the codebase. Additionally, there are improvements to authentication handling for micro apps and minor fixes in service and UI components.

**Micro App Mode Refactor:**

* Removed the `useMicroApp` hook and replaced all usages with the `isMicroApp` boolean from `@config/config`, simplifying how micro app mode is checked across the app (`App.tsx`, `AppHandler.tsx`, `Layout.tsx`, `BaseQuery.ts`, `config.ts`). [[1]](diffhunk://#diff-887196847ba9f7312f9a31db5b0de9a93565367671485d007fe064c07d065053L24-L28) [[2]](diffhunk://#diff-887196847ba9f7312f9a31db5b0de9a93565367671485d007fe064c07d065053L76-R77) [[3]](diffhunk://#diff-fa4e8fe2ae3df2a172861aaf28da11847c2ef0572dfa8bcd5ff6d72038823ac4L23-R30) [[4]](diffhunk://#diff-fa4e8fe2ae3df2a172861aaf28da11847c2ef0572dfa8bcd5ff6d72038823ac4L41) [[5]](diffhunk://#diff-fa4e8fe2ae3df2a172861aaf28da11847c2ef0572dfa8bcd5ff6d72038823ac4L51-R62) [[6]](diffhunk://#diff-d5f8ed36e79e9ae2b29f958a73d64c8c4c75d09aaee6aaa0806b568f55a31366L27-R31) [[7]](diffhunk://#diff-d5f8ed36e79e9ae2b29f958a73d64c8c4c75d09aaee6aaa0806b568f55a31366L44) [[8]](diffhunk://#diff-d5f8ed36e79e9ae2b29f958a73d64c8c4c75d09aaee6aaa0806b568f55a31366L98-R102) [[9]](diffhunk://#diff-d5f8ed36e79e9ae2b29f958a73d64c8c4c75d09aaee6aaa0806b568f55a31366L155-R154) [[10]](diffhunk://#diff-a0988f3541c5041b90601c781278c19e6769c968055b783fd3790538533c36b8L20) [[11]](diffhunk://#diff-a0988f3541c5041b90601c781278c19e6769c968055b783fd3790538533c36b8L95-R96) [[12]](diffhunk://#diff-5da89998b8af962c6dc37cd2a1e50a64b6551da9fecb58696d8b7bacc8d9020fL16) [[13]](diffhunk://#diff-5da89998b8af962c6dc37cd2a1e50a64b6551da9fecb58696d8b7bacc8d9020fL29-R28) [[14]](diffhunk://#diff-5da89998b8af962c6dc37cd2a1e50a64b6551da9fecb58696d8b7bacc8d9020fL54-R53) [[15]](diffhunk://#diff-36742785a17c39684c200e9702fc7c4df6ac02e05831a4754ac5b63e825bf9c9L1-L28)

**Authentication and Privileges Handling:**

* Updated `MicroAppAuthProvider` to fetch user info and privileges on authentication, improving error handling and ensuring necessary data is loaded for micro apps (`AuthContext.tsx`).

**Service Layer Improvements:**

* Added the `x-jwt-assertion` header in API requests when an access token is present, ensuring proper authentication headers for backend services (`BaseQuery.ts`).
* Simplified retry backoff logic in API requests by removing unnecessary parameters (`BaseQuery.ts`).

**UI and Unused Code Cleanup:**

* Removed unused imports and code related to Lottie animations from the `DinnerOnDemand` component, cleaning up the UI code (`DinnerOnDemand.tsx`). [[1]](diffhunk://#diff-eb5b0946cd0a524fe90ace9f8004bce4646716b2fcb5c824e1f0db244732c26fL19-L26) [[2]](diffhunk://#diff-eb5b0946cd0a524fe90ace9f8004bce4646716b2fcb5c824e1f0db244732c26fL54-L62) [[3]](diffhunk://#diff-eb5b0946cd0a524fe90ace9f8004bce4646716b2fcb5c824e1f0db244732c26fR63-L82)

These changes make the codebase cleaner and more maintainable, especially regarding how micro app mode is handled and how authentication is managed for micro apps.

**UIs:**
Screenshot 2026-05-25 at 17 45 50
Screenshot 2026-05-25 at 17 45 59

## Summary by CodeRabbit

* **New Features**
  * Enhanced authentication: fetches user info and loads privileges automatically after login.

* **UI Changes**
  * Removed animated logo from Dinner on Demand empty state.
  * Updated DinnerOnDemand availability text (“till” → “to”).
  * Home now shows Menu and DinnerOnDemand together; layout hides header/sidebar and adjusts mobile bottom bar for micro-app mode.

* **Refactor**
  * Micro-app detection standardized to a configuration flag.

* **Chores**
  * Removed unused animation asset, deprecated micro-app hook, and the MicroAppType enum.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wso2-open-operations/people-ops-suite/pull/280?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


apps/menu-app/webapp/src/context/AuthContext.tsx
```diff
@@ -218,23 +218,31 @@ const AppAuthProvider = (props: { children: React.ReactNode }) => {
 const MicroAppAuthProvider = (props: { children: React.ReactNode }) => {
   const [appState, setAppState] = useState<AppState>(AppState.Loading);
   let mounted = true;
```
chanukaranaba: `let mounted = true;` is declared at component scope, so it's **re-initialized to `true` on every render**. The cleanup in the `useEffect` return sets it to `false`, but the next render immediately resets it — so the `if (!mounted) return;` guard inside the `getToken` callback isn't reliably doing what it looks like it's doing.

Pre-existing, but worth fixing while you're touching this provider. Two safe patterns:

```tsx
// either: ref
const mountedRef = useRef(true);
useEffect(() => {
  return () => { mountedRef.current = false; };
}, []);

// or: scope the flag inside the effect
useEffect(() => {
  let mounted = true;
  // ...
  return () => { mounted = false; };
}, []);
```

chanukaranaba (commented):
Nice cleanup overall — dropping `useMicroApp` in favor of the module-level `isMicroApp` removes a bit of indirection, and ditching the 1498-line Lottie JSON + the recolor hook is a solid bundle-size win. The mechanical `useMicroApp()` → `isMicroApp` swaps in `App.tsx`, `AppHandler.tsx`, and `Layout.tsx` read fine.

A few things worth a second look before merging:

1. **Semantic change**: the old `useMicroApp` did `Object.values(MicroAppType).includes(isMicroApp as MicroAppType)` — a strict enum-membership check. The new code uses the raw value as a boolean, so any truthy string (e.g. `"false"`, or any non-empty string left over from an old HTML template) will now flip the app into micro-app mode. Confirm whatever populates `window.config.IS_MICROAPP` actually produces a real boolean.
2. **`MicroAppAuthProvider` race**: `setAppState(Authenticated)` now fires before `setupAuthenticatedUser()` finishes loading user info / privileges, so any consumer that reads roles or user info immediately on `Authenticated` may see them empty for a tick. Also worth checking the `useEffect` deps array (not visible in the hunk) — it likely needs `triggerGetUserInfo` and `dispatch`.
3. **PR description vs diff mismatch**: the description mentions adding an `x-jwt-assertion` header in `BaseQuery.ts`, but I don't see that change in the diff. Either it was dropped or it lives in another commit/file — worth double-checking.

Inline notes below.

<sub>_Reviewed by Claude (posted via @chanukaranaba)._</sub>

apps/menu-app/webapp/src/context/AuthContext.tsx
```diff
@@ -218,23 +218,31 @@ const AppAuthProvider = (props: { children: React.ReactNode }) => {
...
     };
 
-    getIdToken();
+    getToken((token) => {
+      if (!mounted) return;
+
+      if (token) {
+        setTokens(token, null, null);
+        setAppState(AppState.Authenticated);
+        setupAuthenticatedUser();
```
chanukaranaba: Three things tangled together on this line:

1. **State race.** `setAppState(AppState.Authenticated)` runs *before* `setupAuthenticatedUser()` resolves, so the rest of the app sees `Authenticated` while `userInfo` and `privileges` are still in-flight. If any downstream code reads roles/user info on the `Authenticated` transition, it'll briefly see them empty. Consider awaiting `setupAuthenticatedUser()` before flipping state, or introducing an intermediate `LoadingProfile` state.
2. **Fire-and-forget rejection.** `setupAuthenticatedUser()` is not awaited; if `dispatch(loadPrivileges())` throws, you get an unhandled promise rejection. Wrap the call in `.catch(...)` or `await` it inside an `async` IIFE.
3. **No `mounted` re-check after the awaits inside `setupAuthenticatedUser`.** If the component unmounts during `triggerGetUserInfo` or `loadPrivileges`, the `dispatch` calls still fire on the (now stale) store and React will warn about state updates on unmounted components.

apps/menu-app/webapp/src/context/AuthContext.tsx
```diff
@@ -218,23 +218,31 @@ const AppAuthProvider = (props: { children: React.ReactNode }) => {
...
-        }
-      });
+    const setupAuthenticatedUser = async () => {
+      const userInfoResult = await triggerGetUserInfo();
+      if (userInfoResult?.isError) {
+        console.error("Failed to fetch user info:", userInfoResult.error);
+        dispatch(setAuthError());
+      }
+      await dispatch(loadPrivileges());
     };
```
chanukaranaba, line 241: Minor: `setupAuthenticatedUser` only handles the `isError` branch — if the underlying RTK Query throws (network error before the response is shaped into a result) or `loadPrivileges` rejects, this `async` function rejects silently because it's called without `await` or `.catch` at the call site (line 240). Add a `try/catch` here to make failures observable.

apps/menu-app/webapp/src/config/config.ts
```diff
@@ -26,7 +25,7 @@ declare global {
       AUTH_SIGN_IN_REDIRECT_URL: string;
       AUTH_SIGN_OUT_REDIRECT_URL: string;
       REACT_APP_BACKEND_BASE_URL: string;
-      IS_MICROAPP: MicroAppType;
+      IS_MICROAPP: boolean;
```
chanukaranaba, line 28: Changing `IS_MICROAPP: MicroAppType` → `boolean` tightens the type, but the actual value still comes from `window.config` — typically populated by an HTML template, build-time inject, or env substitution.

If that injection currently writes a string (`"true"` / `"false"`) instead of a real boolean, TypeScript won't catch it (the type is a lie at runtime) and downstream truthy checks like `isMicroApp ? ...` will flip the app into micro-app mode for **any** non-empty string, including `"false"`. Worth grep'ing for the IS_MICROAPP substitution point (likely `index.html` or a config template) and confirming it emits a JS boolean literal, not a quoted string.

apps/menu-app/webapp/src/config/config.ts
```diff
@@ -51,4 +50,4 @@ export const AppConfig = {
   },
 };
 
-export const isMicroApp: MicroAppType = window.config.IS_MICROAPP;
+export const isMicroApp = window.config.IS_MICROAPP;
```
chanukaranaba, line 53: Semantic change worth being explicit about in the PR description: previously consumers checked `Object.values(MicroAppType).includes(isMicroApp as MicroAppType)` — i.e. only flipped to micro-app mode for *known enum values*. After this PR, the raw value is used as a boolean.

If any deployment ever has `window.config.IS_MICROAPP` set to something other than a strict boolean (`""`, `"false"`, `0`, `null`, etc.), the behavior diverges from before. Either:
- Guarantee the host writes a real boolean (tied to the comment on line 28), or
- Coerce here: `export const isMicroApp = window.config.IS_MICROAPP === true;` to keep the strict semantics.

apps/menu-app/webapp/src/services/BaseQuery.ts
```diff
@@ -92,9 +91,8 @@ export const baseQueryWithReauth: BaseQueryFn<
  */
 export const baseQueryWithRetry = retry(
   async (args: string | FetchArgs, api, extraOptions) => {
-    const isValidMicroApp = Object.values(MicroAppType).includes(isMicroApp as MicroAppType);
 
-    const result = isValidMicroApp
+    const result = isMicroApp
```
chanukaranaba, line 95: Same semantic concern as in `config.ts` — the previous `isValidMicroApp` check was enum-membership-based; this is now a raw truthy check. If `window.config.IS_MICROAPP` is ever a non-boolean (string from HTML injection, etc.), this branch goes through `baseQuery` (no reauth) instead of `baseQueryWithReauth`, which is a meaningful auth-flow difference.

If the coercion is added at the config layer (see comment on `config.ts:53`), this site becomes safe automatically — no change needed here.

apps/menu-app/webapp/src/context/AuthContext.tsx
dinethde: Fixed in the commit https://github.com/wso2-open-operations/people-ops-suite/pull/280/changes/ea0879a5d71dc79e999fa4e07b7edd12ac7b6413

apps/menu-app/webapp/src/context/AuthContext.tsx
dinethde: Fixed in the commit [9b50263](https://github.com/wso2-open-operations/people-ops-suite/pull/280/commits/9b50263b9cba067d75f466bbe6b3823ba47b285c)

apps/menu-app/webapp/src/context/AuthContext.tsx
dinethde: Fixed in the commit [9b50263](https://github.com/wso2-open-operations/people-ops-suite/pull/280/commits/9b50263b9cba067d75f466bbe6b3823ba47b285c)

apps/menu-app/webapp/src/config/config.ts
dinethde: Actual value is a boolean.

apps/menu-app/webapp/src/config/config.ts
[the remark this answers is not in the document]
dinethde: Actual value is now a Boolean value

apps/menu-app/webapp/src/context/AuthContext.tsx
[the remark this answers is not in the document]
dinethde: FIxed in the commit [ea0879a](https://github.com/wso2-open-operations/people-ops-suite/pull/280/commits/ea0879a5d71dc79e999fa4e07b7edd12ac7b6413)

apps/menu-app/webapp/src/context/AuthContext.tsx
```diff
@@ -218,23 +218,31 @@ const AppAuthProvider = (props: { children: React.ReactNode }) => {
...
+      if (!mounted) return;
+
+      if (token) {
+        setTokens(token, null, null);
+        setAppState(AppState.Authenticated);
+        setupAuthenticatedUser();
+      } else {
+        setAppState(AppState.Unauthenticated);
+        console.error("No token available");
+      }
```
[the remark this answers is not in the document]
dinethde: Fixed in the commit [ea0879a](https://github.com/wso2-open-operations/people-ops-suite/pull/280/commits/ea0879a5d71dc79e999fa4e07b7edd12ac7b6413)

apps/menu-app/webapp/src/context/AuthContext.tsx
```diff
@@ -218,23 +218,31 @@ const AppAuthProvider = (props: { children: React.ReactNode }) => {
...
+      await dispatch(loadPrivileges());
     };
 
-    getIdToken();
+    getToken((token) => {
+      if (!mounted) return;
+
+      if (token) {
+        setTokens(token, null, null);
```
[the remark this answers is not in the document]
dinethde: -1. In the code level, we send the Authorization header with the bearer schema, and the gateway is responsible to add x-jwt-assertion header

apps/menu-app/webapp/src/config/config.ts
```diff
@@ -26,7 +25,7 @@ declare global {
       AUTH_SIGN_IN_REDIRECT_URL: string;
       AUTH_SIGN_OUT_REDIRECT_URL: string;
       REACT_APP_BACKEND_BASE_URL: string;
-      IS_MICROAPP: MicroAppType;
+      IS_MICROAPP: boolean;
     };
```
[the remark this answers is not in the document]
dinethde: -1

apps/menu-app/webapp/src/context/AuthContext.tsx
```diff
@@ -218,23 +218,31 @@ const AppAuthProvider = (props: { children: React.ReactNode }) => {
...
-          setAppState(AppState.Authenticated);
-        } else {
-          setAppState(AppState.Unauthenticated);
-          console.error("No token available");
-        }
-      });
+    const setupAuthenticatedUser = async () => {
+      const userInfoResult = await triggerGetUserInfo();
+      if (userInfoResult?.isError) {
+        console.error("Failed to fetch user info:", userInfoResult.error);
```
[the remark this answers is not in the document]
dinethde: Fixed

apps/menu-app/webapp/src/services/BaseQuery.ts
```diff
@@ -92,9 +91,8 @@ export const baseQueryWithReauth: BaseQueryFn<
...
       ? await baseQuery(args, api, extraOptions)
       : await baseQueryWithReauth(args, api, extraOptions);
```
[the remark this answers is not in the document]
dinethde: fixed

sacheeramesh (approved):

PR #294, opened 2026-06-12
[People App] Employee onboarding fixes

shayanmalinda:
## Summary
Fixed employee-ID generation for country-coded company prefixes, restructured the onboarding navigation, filtered inactive employment types, and disabled Asgardeo provisioning in bulk onboarding (to be re-enabled later).

## What changed

1. Employee-ID generation (backend)
2. Company prefix backfill (DB migration)
3. Filter inactive employment types (full stack)
4. Bulk onboarding → nested sidenav (frontend)
5. Disable Asgardeo provisioning in bulk onboarding (backend)

## Summary by CodeRabbit

* **New Features**
  * Bulk onboarding added as a dedicated view and route; onboarding now has separate single and bulk pages
  * "Probation" added as an employment type; employment-type data now includes active/inactive status

* **Bug Fixes**
  * Employee ID generation trims/validates company prefixes and rejects inactive employment types
  * Employment type selector shows only active options (but preserves current selection)

* **Chores**
  * Backend dependency patch updates
  * Database backfill script to populate company prefixes
* **Behavior Change**
  * External provisioning/group-assignment flows disabled; responses include empty provisioning arrays

### DIFF
--- a/apps/people-app/backend/modules/database/db_functions.bal
+++ b/apps/people-app/backend/modules/database/db_functions.bal
@@ -498,10 +498,18 @@ isolated function generateBulkEmployeeId(CreateEmployeePayload payload,
 
     match context.employmentType {
         PERMANENT|INTERNSHIP => {
+            if context.companyPrefix.trim().length() == 0 {
+                return error(string `Company (ID: ${payload.companyId}) has no employee ID prefix configured`);
+            }
+            // PERMANENT and PROBATION share one number line; INTERNSHIP runs a separate one.
+            // Scope both the MAX query and the in-batch cache key to the matching group.
+            EmploymentTypeName[] sequenceTypes = context.employmentType == PERMANENT
+                ? [PERMANENT, PROBATION]
+                : [INTERNSHIP];
             string seqKey = context.companyPrefix + ":" + context.employmentType.toString();
             if !sequenceCache.hasKey(seqKey) {
                 EmployeeIdSequence seq = check getLastEmployeeNumericSuffix(
-                        context.companyPrefix, [context.employmentType]);
+                        context.companyPrefix, sequenceTypes);
                 sequenceCache[seqKey] = <int>seq.lastNumericId;
             }
             int next = (sequenceCache[seqKey] ?: 0) + 1;
--- a/apps/people-app/backend/modules/database/db_functions.bal
+++ b/apps/people-app/backend/modules/database/db_functions.bal
@@ -498,23 +498,27 @@ isolated function generateBulkEmployeeId(CreateEmployeePayload payload,
 
     match context.employmentType {
         PERMANENT|INTERNSHIP => {
-            if context.companyPrefix.trim().length() == 0 {
+            // Normalize the prefix once so a value like " SG " can't pass the guard, leak spaces
+            // into the ID, or fork a separate sequence from "SG".
+            string companyPrefix = context.companyPrefix.trim();
+            if companyPrefix.length() == 0 {
                 return error(string `Company (ID: ${payload.companyId}) has no employee ID prefix configured`);
             }
             // PERMANENT and PROBATION share one number line; INTERNSHIP runs a separate one.
-            // Scope both the MAX query and the in-batch cache key to the matching group.
+            // Scope both the MAX query and the in-batch cache key to the matching group. PROBATION
+            // never reaches this path (inactive types are rejected during bulk validation).
             EmploymentTypeName[] sequenceTypes = context.employmentType == PERMANENT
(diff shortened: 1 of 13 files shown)


apps/people-app/backend/utils.bal
```diff
@@ -43,8 +43,24 @@ public isolated function generateEmployeeId(database:CreateEmployeePayload paylo
 
     match ctx.employmentType {
         database:PERMANENT|database:INTERNSHIP => {
```
chanukaranaba, line 63: **Bug — PROBATION falls through to `Unsupported employment type` (HTTP 500).** This arm matches only `PERMANENT|INTERNSHIP`, and there's no PROBATION arm below — so a PROBATION employment type hits the `_ =>` default and returns a 500. Since `JobInfo.tsx` now lists all *active* employment types, selecting an active PROBATION type breaks onboarding.

Note the `sequenceTypes` ternary's PROBATION branch is currently **unreachable**: inside this arm `employmentType` is only ever PERMANENT or INTERNSHIP. Fix the arm to include PROBATION and key the ternary off INTERNSHIP:
```ballerina
database:PERMANENT|database:PROBATION|database:INTERNSHIP => {
    ...
    database:EmploymentTypeName[] sequenceTypes = ctx.employmentType == database:INTERNSHIP
        ? [database:INTERNSHIP]
        : [database:PERMANENT, database:PROBATION];
```
That's the only path that actually gives a PROBATION hire a `<prefix><number>` ID, and it matches the stated "keep the same ID across probation → permanent" intent.

apps/people-app/backend/modules/database/db_functions.bal
```diff
@@ -498,10 +498,18 @@ isolated function generateBulkEmployeeId(CreateEmployeePayload payload,
 
     match context.employmentType {
         PERMANENT|INTERNSHIP => {
```
chanukaranaba, line 500: **Same PROBATION gap in the bulk path.** This arm is `PERMANENT|INTERNSHIP` with no PROBATION arm, so a bulk PROBATION row falls to `_ =>` → `error("Unsupported employment type: PROBATION")`. Mirror the single-onboarding fix: match `PERMANENT|PROBATION|INTERNSHIP` and set `sequenceTypes = context.employmentType == INTERNSHIP ? [INTERNSHIP] : [PERMANENT, PROBATION]`.

Also double-check the `seqKey` (`companyPrefix + ":" + employmentType`) — with the fix, PERMANENT and PROBATION must share the **same** cache key (since they share one number line), otherwise two in-batch sequences could hand out the same number. Key it on the sequence group, not the raw employment type.

apps/people-app/backend/service.bal
```diff
@@ -16,9 +16,9 @@
 
 import people.authorization;
 import people.database;
-import people.email;
+// import people.email;
 import people.qr;
-import people.scim;
+// import people.scim;
```
chanukaranaba: **Commented-out code (maintainability).** The bulk Asgardeo/SCIM provisioning block (~163 lines) plus these `// import people.email;` / `// import people.scim;` lines are commented rather than removed. It's compile-safe — there are no live `scim:`/`email:` references left anywhere in the backend — but a large commented block "to be re-enabled later" tends to rot and obscures the diff. Prefer removing it (git history preserves it for restoration) or gating the live block behind a config/feature flag so it can be toggled without a code change.

chanukaranaba (commented):
Reviewed the onboarding fixes. The empty-prefix guard and the shared PERMANENT/PROBATION number-line reasoning are solid, and the is_active filtering is threaded cleanly full-stack. One real bug — PROBATION isn't actually handled by the ID-generation match — plus a maintainability note, inline.

apps/people-app/backend/utils.bal
shayanmalinda: It's not actually a BUG, `PROBATION` is not a valid employment type anymore in the system. But there are historical employees under that employment type, which we have to keep it as a inactive employment type.

Additional active employment type check has been introduced to throw Bad Request, when a user is onboarding with an inactive employment type.

apps/people-app/backend/modules/database/db_functions.bal
shayanmalinda: DITTO

apps/people-app/backend/service.bal
shayanmalinda: This code block needs to be uncommented once the Asgardeo provisioning configuration and testing are completed. Until then, it will remain temporarily commented out. A comment has been added to explain this, so that anyone referring to the source code in the future will understand the reason.

chanukaranaba (approved):

PR #306, opened 2026-07-17
[Visitor Microapp] Bugfix "Whom They Meet" self-prepopulation for external logins and cleared field

v15a1:
## Summary
- `decodeTokenAndStoreUser` (auth.ts) only read `given_name`/`family_name` from the ID token, ignoring the `name` claim. When both were absent (common for externally-federated logins where the IdP only asserts a combined `name`), the name became a single whitespace character, which is truthy — so downstream `user.name || user.email` fallbacks never kicked in and the "Whom They Meet" field appeared blank instead of showing the visitor host's own email.
- Hardened both prepopulation fallback sites in CreateVisit.tsx to `user.name?.trim() || user.email` so a future empty/whitespace name can't silently win again.
- Separately, once "Whom They Meet" was cleared mid-session, nothing ever offered the logged-in user back as an option — the only prepopulation logic was a mount-time effect keyed on `user`, which doesn't re-fire just because the field emptied. Fixed by pinning the current user as the first Autocomplete option whenever the search input is empty.

## Test plan
- [x] `tsc --noEmit` passes
- [x] `eslint` passes (no new warnings)
- [x] Manual verification in the running app: log in with an externally-federated account, confirm the field prepopulates with the user's name/email; clear the field and reopen the dropdown, confirm "myself" appears as the first option

Closes https://github.com/wso2-enterprise/wso2-digital-team-project-management/issues/674

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the “Whom They Meet” field so the signed-in user stays available as a selectable option after clearing.
  * Improved name handling when creating visits and initializing user information (better trimming and claim-based fallback).
  * Enhanced employee lookup and authentication failure handling to prevent missing data and improve resilience.

* **Logging**
  * Added richer runtime diagnostics for token retrieval/decoding, user initialization, employee lookups, and visit form prepopulation/reset flows to aid troubleshooting.

### DIFF
--- a/apps/visitor-app/micro-app/src/microapp-bridge/index.ts
+++ b/apps/visitor-app/micro-app/src/microapp-bridge/index.ts
@@ -21,7 +21,7 @@ import { BrowserConfiguration, Topic } from "./types";
  * Internal log helper to avoid circular dependency with Logger.
  * Logger imports sendNativeLog from this module, so we cannot import Logger here.
  */
-const bridgeLog = (message: string, level: "info" | "error" | "warn" = "error") => {
+export const bridgeLog = (message: string, level: "info" | "error" | "warn" = "error") => {
   if (typeof window !== "undefined" && window.ReactNativeWebView) {
     window.ReactNativeWebView.postMessage(
       JSON.stringify({ topic: Topic.nativeLog, data: { message, level } })
@@ -36,8 +36,13 @@ type Callback<T> = (data?: T) => void;
 // Function to get token from React Native
 export const getToken = (callback: Callback<string>): void => {
   if (window.nativebridge) {
+    bridgeLog("getToken: requesting token from native bridge", "info");
     window.nativebridge.requestToken();
     window.nativebridge.resolveToken = (token: string) => {
+      bridgeLog(
+        `getToken: resolved token (present=${!!token}, length=${token?.length ?? 0})`,
+        "info",
+      );
       callback(token);
     };
   } else {
--- a/apps/visitor-app/micro-app/src/microapp-bridge/index.ts
+++ b/apps/visitor-app/micro-app/src/microapp-bridge/index.ts
@@ -36,13 +36,8 @@ type Callback<T> = (data?: T) => void;
 // Function to get token from React Native
 export const getToken = (callback: Callback<string>): void => {
   if (window.nativebridge) {
-    bridgeLog("getToken: requesting token from native bridge", "info");
     window.nativebridge.requestToken();
     window.nativebridge.resolveToken = (token: string) => {
-      bridgeLog(
-        `getToken: resolved token (present=${!!token}, length=${token?.length ?? 0})`,
-        "info",
-      );
       callback(token);
     };
   } else {
--- a/apps/visitor-app/micro-app/src/pages/CreateVisit/CreateVisit.tsx
+++ b/apps/visitor-app/micro-app/src/pages/CreateVisit/CreateVisit.tsx
@@ -325,10 +325,6 @@ function CreateVisit() {
   // Pre-populate "Whom They Meet" with the logged-in user's details
   useEffect(() => {
     if (user?.email && !whoTheyMeet) {
-      bridgeLog(
-        `CreateVisit: prepopulating whoTheyMeet (name=${!!user.name})`,
-        "info",
-      );
       const nameParts = (user.name || "").trim().split(/\s+/).filter(Boolean);
       const firstName = nameParts[0] || user.email;
       const lastName = nameParts.slice(1).join(" ");
@@ -354,10 +350,6 @@ function CreateVisit() {
             offset: 0,
             limit: 1,
           });
-          bridgeLog(
-            `CreateVisit: fetching employee record for ${user.email}`,
-            "info",
-          );
           const res = await apiClient.get(
             `${endpoint.baseUrl}${endpoint.path}`,
           );
@@ -366,7 +358,6 @@ function CreateVisit() {
             (emp) => emp.workEmail.toLowerCase() === user.email.toLowerCase(),
           );
           if (match) {
-            bridgeLog("CreateVisit: employee match found", "info");
             setSelectedEmployee(match);
             const empFullName =
               `${match.firstName || ""} ${match.lastName || ""}`.trim();
@@ -705,10 +696,6 @@ function CreateVisit() {
 
         // Pre-populate "Whom They Meet" with the logged-in user
         if (user?.email) {
-          bridgeLog(
-            `CreateVisit.resetForm: prepopulating whoTheyMeet (name=${!!user.name})`,
-            "info",
-          );
           const nameParts = (user.name || "").trim().split(/\s+/).filter(Boolean);
           const firstName = nameParts[0] || user.email;
           const lastName = nameParts.slice(1).join(" ");
@@ -733,10 +720,6 @@ function CreateVisit() {
                 offset: 0,
                 limit: 1,
               });
-              bridgeLog(
-                `CreateVisit.resetForm: fetching employee record for ${user.email}`,
-                "info",
-              );
               const res = await apiClient.get(
                 `${endpoint.baseUrl}${endpoint.path}`,
               );
@@ -746,7 +729,6 @@ function CreateVisit() {
                   emp.workEmail.toLowerCase() === user.email.toLowerCase(),
               );
               if (match) {
-                bridgeLog("CreateVisit.resetForm: employee match found", "info");
                 setSelectedEmployee(match);
                 const empFullName =
                   `${match.firstName || ""} ${match.lastName || ""}`.trim();
--- a/apps/visitor-app/micro-app/src/pages/CreateVisit/CreateVisit.tsx
+++ b/apps/visitor-app/micro-app/src/pages/CreateVisit/CreateVisit.tsx
@@ -28,6 +28,7 @@ import QRCode from "qrcode";
 import { PhoneNumberUtil } from "google-libphonenumber";
 import { MuiTelInput } from "mui-tel-input";
 import { hash } from "../../utils/utils";
+import { bridgeLog } from "../../microapp-bridge";
 import { useSnackbar } from "../../contexts/SnackbarContext";
 import { useDialog } from "../../contexts/DialogContext";
 import { useUserStore } from "../../stores/user/user";
@@ -324,10 +325,14 @@ function CreateVisit() {
   // Pre-populate "Whom They Meet" with the logged-in user's details
   useEffect(() => {
     if (user?.email && !whoTheyMeet) {
+      bridgeLog(
+        `CreateVisit: prepopulating whoTheyMeet from user ${JSON.stringify(user)}`,
+        "info",
+      );
       const nameParts = (user.name || "").trim().split(/\s+/);
       const firstName = nameParts[0] || "";
       const lastName = nameParts.slice(1).join(" ") || "";
-      const fullName = user.name || user.email;
+      const fullName = user.name?.trim() || user.email;
 
       setWhoTheyMeet(user.email);
       setWhoTheyMeetName(fullName);
@@ -349,14 +354,26 @@ function CreateVisit() {
             offset: 0,
             limit: 1,
           });
+          bridgeLog(
+            `CreateVisit: fetching employee record for ${user.email}`,
+            "info",
+          );
           const res = await apiClient.get(
             `${endpoint.baseUrl}${endpoint.path}`,
+          );
+          bridgeLog(
+            `CreateVisit: employee search response ${JSON.stringify(res.data)}`,
+            "info",
           );
           const employees: Employee[] = res.data;
           const match = employees?.find(
             (emp) => emp.workEmail.toLowerCase() === user.email.toLowerCase(),
           );
           if (match) {
+            bridgeLog(
+              `CreateVisit: employee match found ${JSON.stringify(match)}`,
+              "info",
+            );
             setSelectedEmployee(match);
             const empFullName =
               `${match.firstName || ""} ${match.lastName || ""}`.trim();
@@ -364,9 +381,15 @@ function CreateVisit() {
               setWhoTheyMeetName(empFullName);
               setInputValue(empFullName);
             }
+          } else {
+            bridgeLog(
+              `CreateVisit: no employee match found for ${user.email}`,
+              "warn",
+            );
           }
-        } catch {
+        } catch (err) {
           // Thumbnail fetch failed silently — fallback already set above
+          bridgeLog(`CreateVisit: employee lookup failed - ${err}`, "error");
         }
       })();
     }
@@ -390,7 +413,28 @@ function CreateVisit() {
     true,
   );
 
-  const employees = employeesData || [];
+  // Pin the logged-in user as the first option once the field is cleared
+  const selfEmployeeOption: Employee | null = useMemo(() => {
+    if (!user?.email) return null;
+    const nameParts = (user.name || "").trim().split(/\s+/).filter(Boolean);
+    return {
+      firstName: nameParts[0] || user.email,
+      lastName: nameParts.slice(1).join(" "),
+      workEmail: user.email,
+      employeeThumbnail: null,
+    };
+  }, [user]);
+
+  const employees = useMemo(() => {
+    const base = employeesData || [];
+    if (inputValue.trim() || !selfEmployeeOption) return base;
+    const alreadyIncluded = base.some(
+      (emp) =>
+        emp.workEmail.toLowerCase() ===
+        selfEmployeeOption.workEmail.toLowerCase(),
+    );
+    return alreadyIncluded ? base : [selfEmployeeOption, ...base];
+  }, [employeesData, inputValue, selfEmployeeOption]);
 
   // Add visitor mutation
   const addVisitorMutation = useAPI<any, AddVisitorPayload>(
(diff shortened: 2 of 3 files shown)


shayanmalinda (approved):

apps/visitor-app/micro-app/src/pages/CreateVisit/CreateVisit.tsx
```diff
@@ -324,10 +325,10 @@ function CreateVisit() {
...
-      const firstName = nameParts[0] || "";
-      const lastName = nameParts.slice(1).join(" ") || "";
-      const fullName = user.name || user.email;
+      const nameParts = (user.name || "").trim().split(/\s+/).filter(Boolean);
+      const firstName = nameParts[0] || user.email;
+      const lastName = nameParts.slice(1).join(" ");
+      const fullName = user.name?.trim() || user.email;
 
       setWhoTheyMeet(user.email);
       setWhoTheyMeetName(fullName);
```
v15a1, line 334: **This is not a mount effect. It re-runs on every API request and undoes the clear.**

The dep array is `[user]` (L380), and `user` gets a fresh object identity on every request:

- `apiClient`'s request interceptor calls `refreshToken()` before every request (`services/apiClient.ts:45-49`).
- That calls `initializeUserFromToken` → `decodeTokenAndStoreUser` → `setUser({ email, name })`, a new object literal each time (`services/auth.ts:87-91`).
- The zustand selector on L150 compares by reference, so `CreateVisit` re-renders and this effect re-runs.

The guard is `!whoTheyMeet`, which is precisely the state the new pinned-option feature creates. Repro:

1. Field is prepopulated with self.
2. Clear it (X button, or backspace to empty; MUI nulls the value on both paths).
3. Type `bob`. At 500ms the debounce fires the employee search.
4. That request triggers the token refresh → `setUser` → this effect runs → `setInputValue(<your own name>)`, `setSelectedEmployee(self)`, `setWhoTheyMeet(user.email)`.

The typed text is replaced with your own name exactly as the results land, and the effect fires another `/employees` request each time. The pinned-self dropdown option is hard to reach because the field refills itself first.

This predates the PR, but goal (b) depends on it. It also explains why it wasn't caught locally: in a plain browser `window.nativebridge` is absent, `getToken` calls back with `undefined`, `refreshToken` rejects and `setUser` never runs, so the loop only exists inside the real WebView.

The `eslint-disable exhaustive-deps` on L380 is hiding it rather than documenting it: the effect reads `whoTheyMeet` without tracking it. A ref recording "already prepopulated once", or making `setUser` skip the write when the decoded claims are unchanged, would both close it.

apps/visitor-app/micro-app/src/pages/CreateVisit/CreateVisit.tsx
```diff
@@ -364,9 +365,15 @@ function CreateVisit() {
               setWhoTheyMeetName(empFullName);
               setInputValue(empFullName);
             }
+          } else {
+            bridgeLog(
+              `CreateVisit: no employee match found for ${user.email}`,
+              "warn",
+            );
```
v15a1, line 372: **PII survived the scrub.** `${user.email}` goes onto the native log channel here and again at L741. `bridgeLog` posts to `Topic.nativeLog` (`microapp-bridge/index.ts:24-31`), so this leaves the device and lands in the super app's log sink.

`f8168640` ("drop PII from logs") removed `JSON.stringify(user)`, `JSON.stringify(res.data)` and `JSON.stringify(match)`, and even rewrote one message to `(name=${!!user.name})` to avoid leaking the name. These two raw email interpolations were missed. Same CWE-532 class CodeRabbit flagged.

Separately, `warn` is the wrong level. "No employee match" is the normal outcome for anyone outside the employee index (contractors, new joiners), and it fires on every mount, so it is a per-mount warning for expected behaviour.

apps/visitor-app/micro-app/src/pages/CreateVisit/CreateVisit.tsx
```diff
@@ -364,9 +365,15 @@ function CreateVisit() {
...
           }
-        } catch {
+        } catch (err) {
           // Thumbnail fetch failed silently — fallback already set above
+          bridgeLog(`CreateVisit: employee lookup failed - ${err}`, "error");
         }
       })();
```
v15a1, line 378: **This lookup is uncancelled and can leave the displayed host and the submitted host pointing at different people.**

The IIFE writes `selectedEmployee`, `whoTheyMeetName` and `inputValue`, but never `whoTheyMeet`, and it never re-checks that the field still refers to the logged-in user before writing.

On a slow connection: page loads, user immediately clears the field and picks "Bob Smith" (`whoTheyMeet = bob@…`). The in-flight self-lookup then resolves and runs `setSelectedEmployee(selfMatch)` + `setInputValue(<your name>)`. The field shows your name and your avatar, while `submitVisitor` still sends `whomTheyMeet: "bob@…"` (L628). Nothing in the UI reveals the mismatch.

MUI widens the window: `useAutocomplete` re-runs `resetInputValue(null, value, 'reset')` whenever the component is unfocused (`useAutocomplete.js:199-210`), so `setSelectedEmployee` on its own is enough to rewrite the input text.

Capturing the email the request was issued for and bailing if `whoTheyMeet` changed in the meantime fixes this and the `resetForm` copy at L714 together.

apps/visitor-app/micro-app/src/pages/CreateVisit/CreateVisit.tsx
```diff
@@ -390,7 +397,28 @@ function CreateVisit() {
...
+  const selfEmployeeOption: Employee | null = useMemo(() => {
+    if (!user?.email) return null;
+    const nameParts = (user.name || "").trim().split(/\s+/).filter(Boolean);
+    return {
+      firstName: nameParts[0] || user.email,
+      lastName: nameParts.slice(1).join(" "),
+      workEmail: user.email,
+      employeeThumbnail: null,
+    };
+  }, [user]);
```
v15a1, line 410: Two things here.

**The synthetic option displaces a better one.** `employeeThumbnail` is hardcoded `null`, and the memo below filters the real API record for the same person out of the list (L415-419) in favour of this one. So: field is prepopulated and shows the user's photo (fetched by the IIFE at L361) → user clears → reopens → picks the pinned first option → the photo is gone, replaced by a letter avatar, and the display name reverts from the directory name to the token-name split. No lookup re-fires, so it never comes back for the rest of the session. Preferring the API record when `employeesData` already contains it avoids the downgrade.

**Third copy of the name parsing.** `(user.name || "").trim().split(/\s+/).filter(Boolean)` now lives here, at L328 and at L699. CodeRabbit's first round already caught two of those copies drifting apart. A single `buildSelfEmployee(user)` helper makes that structurally impossible instead of a review checklist item.

apps/visitor-app/micro-app/src/pages/CreateVisit/CreateVisit.tsx
```diff
@@ -390,7 +397,28 @@ function CreateVisit() {
...
+  const employees = useMemo(() => {
+    const base = employeesData || [];
+    if (inputValue.trim() || !selfEmployeeOption) return base;
+    const filteredBase = base.filter(
+      (emp) =>
+        emp.workEmail.toLowerCase() !==
+        selfEmployeeOption.workEmail.toLowerCase(),
+    );
+    return [selfEmployeeOption, ...filteredBase];
+  }, [employeesData, inputValue, selfEmployeeOption]);
```
v15a1, line 421: Two adjacent issues about the list this memo feeds.

**1. The `reason` guard never matches.** `onInputChange`'s `if (reason === "reset") return;` (L826) reads as "ignore programmatic updates", but on MUI v7 selection fires `'selectOption'` (`useAutocomplete.js:534,553`), blur fires `'blur'` (L854), and `'reset'` only comes from the value-sync effect (L209). So every time the user picks an employee, `debouncedEmployeeSearch("Bob Smith")` is scheduled and 500ms later fires a wasted `/employees?search=Bob%20Smith` round trip, which per the interceptor also triggers another token refresh. Guarding on `reason !== "input"` is what was intended.

**2. Stale results while the debounce settles.** This memo recomputes synchronously when `inputValue` becomes `""`, but `employeeSearch` only updates 500ms later. Search "bob", clear, reopen immediately, and the list reads `[you, Bob Smith, Bobby Jones, …]`. `isLoadingEmployees` is `isLoading` rather than `isFetching`, so a cached key shows no spinner to signal it. `autoHighlight` keeps index 0 (self) selected, so Enter is still safe; the list content is just misleading.

apps/visitor-app/micro-app/src/pages/CreateVisit/CreateVisit.tsx
```diff
@@ -668,10 +696,10 @@ function CreateVisit() {
...
-          const firstName = nameParts[0] || "";
-          const lastName = nameParts.slice(1).join(" ") || "";
-          const fullName = user.name || user.email;
+          const nameParts = (user.name || "").trim().split(/\s+/).filter(Boolean);
+          const firstName = nameParts[0] || user.email;
+          const lastName = nameParts.slice(1).join(" ");
+          const fullName = user.name?.trim() || user.email;
 
           setWhoTheyMeet(user.email);
           setWhoTheyMeetName(fullName);
```
v15a1, line 705: L697-751 is a near-verbatim copy of L327-378, around 55 lines. The two have already diverged (the `!whoTheyMeet` guard exists only in the effect, the log prefixes differ), and this PR had to apply the same `.filter(Boolean)` edit to both. The next change that touches only one gives the app different behaviour at load than after "Complete Visit".

Also, `resetForm` doesn't clear `employeeSearch` or `autocompleteOpen`, so the fresh form's dropdown still holds the previous visit's search results.

apps/visitor-app/micro-app/src/pages/CreateVisit/CreateVisit.tsx
```diff
@@ -708,9 +736,17 @@ function CreateVisit() {
                   setWhoTheyMeetName(empFullName);
                   setInputValue(empFullName);
                 }
+              } else {
+                bridgeLog(
+                  `CreateVisit.resetForm: no employee match found for ${user.email}`,
+                  "warn",
+                );
```
v15a1, line 743: Second site of the same leak as L369-372: `${user.email}` on the native log channel, at `warn` level for what is an expected outcome.

apps/visitor-app/micro-app/src/services/auth.ts
```diff
@@ -82,20 +81,20 @@ export const decodeTokenAndStoreUser = (): User | null => {
...
-      email: decoded.email,
-      name: decoded.name,
-    });
 
+    const nameFromParts =
+      `${decoded.given_name || ""} ${decoded.family_name || ""}`.trim();
     const user: User = {
       email: decoded.email || "",
-      name: `${decoded.given_name || ""} ${decoded.family_name || ""}`,
+      name: decoded.name || nameFromParts,
```
v15a1, line 89: **`decoded.name` isn't trimmed, which reopens the bug this PR exists to close.**

An IdP that sends `name: "  "` alongside valid `given_name`/`family_name` stores `name: "  "` and discards "John Doe". Downstream, `CreateVisit.tsx:331` does `user.name?.trim() || user.email` and renders the raw email address in "Whom They Meet" — the same visible symptom, arrived at through a different claim shape.

`decoded.name?.trim() || nameFromParts` fixes it, and this is the right place for it. Normalising once at the boundary removes the need for `(user.name || "").trim().split(/\s+/).filter(Boolean)` at three separate call sites in `CreateVisit.tsx`. Right now the invalid value is allowed into the store and every consumer has to defend against it.

Related, on L27-28: `TokenPayload` still declares `given_name: string; family_name: string;` as required. `jwtDecode<TokenPayload>` is an unchecked cast so TS will never catch the mismatch, and the entire premise of this fix is that federated IdPs omit those claims. A future `decoded.given_name.trim()` crashes for exactly the users this PR was written for. Both should be optional. `groups` is now read nowhere and could go.

apps/visitor-app/micro-app/src/services/auth.ts
```diff
@@ -82,20 +81,20 @@ export const decodeTokenAndStoreUser = (): User | null => {
...
     };
-
     useUserStore.getState().setUser(user);
-    Logger.info("User information stored in Zustand store", user);
     return user;
   } catch (error) {
+    bridgeLog(
+      `decodeTokenAndStoreUser: failed - ${error}`,
+      "error",
+    );
```
v15a1, line 97: **This catch emits two native log entries for one failure, and neither is complete.**

`bridgeLog` and `Logger.error` both terminate at `window.ReactNativeWebView.postMessage` with `topic: native_log`, in different shapes:

- `{"message":"decodeTokenAndStoreUser: failed - InvalidTokenError: …","level":"error"}` — has the error text, no `data` key at all
- `{"message":"Failed to decode token and store user information","data":{},"level":"error"}` — has the context label, error content gone

The `{}` is the real defect. `sendNativeLog` JSON.stringifies its `data`, and `Error` has no own enumerable properties:

```
$ node -e 'console.log(JSON.stringify({message:"ctx", data:new Error("boom"), level:"error"}))'
{"message":"ctx","data":{},"level":"error"}
```

That has always been true for every `Logger.error(msg, error)` call in the app (`apiClient.ts:60,67,135`, `auth.ts:54,98,115`) — `jwt-decode`'s `InvalidTokenError` included, since it sets `name` on the prototype. I assume that's why `bridgeLog` was reached for, but the workaround isn't written down and leaves the defect in place everywhere else. Teaching `sendNativeLog` to serialise Errors as `{name, message, stack}` would let this line go away entirely.

The two also disagree about when they fire: `bridgeLog` needs only `ReactNativeWebView`, while `sendNativeLog` requires `nativebridge` too and otherwise discards the caller's message in favour of `NATIVE_BRIDGE_NOT_AVAILABLE` (`microapp-bridge/index.ts:117-119`).

apps/visitor-app/micro-app/src/services/auth.ts
```diff
@@ -112,6 +111,7 @@ export const initializeUserFromToken = (): void => {
   try {
     decodeTokenAndStoreUser();
   } catch (error) {
+    bridgeLog(`initializeUserFromToken: failed - ${error}`, "error");
```
v15a1, line 114: **This catch can't fire, and the failure it was meant to cover is silent.**

`decodeTokenAndStoreUser` wraps its whole body in try/catch and returns `null` rather than rethrowing (L74-101), so the only way to reach here is if the catch block itself throws. `setLoading(true)` on L109 sits outside the try, so it isn't covered either. The same argument makes `refreshToken`'s try/catch on L51-58 dead, along with its `Logger.warn`.

The actual failure mode has no handler at all: on a malformed token, `decodeTokenAndStoreUser` clears the user store and returns `null`, this function returns normally, and `refreshToken` **resolves** with the token as though nothing went wrong (L60). Since the request interceptor calls `refreshToken()` on every request, the app carries on with `user === null` and no caller ever learns. Checking the `null` return here, or having `decodeTokenAndStoreUser` return a result the caller must inspect, would make it observable.

apps/visitor-app/micro-app/src/microapp-bridge/index.ts
```diff
@@ -21,7 +21,7 @@ import { BrowserConfiguration, Topic } from "./types";
  * Internal log helper to avoid circular dependency with Logger.
  * Logger imports sendNativeLog from this module, so we cannot import Logger here.
  */
-const bridgeLog = (message: string, level: "info" | "error" | "warn" = "error") => {
+export const bridgeLog = (message: string, level: "info" | "error" | "warn" = "error") => {
```
v15a1, line 24: The doc comment immediately above still says "Internal log helper", and the reason it gives (circular dependency with Logger) only applies inside this module. Neither new caller has that constraint: `auth.ts` already imports `Logger` on L20, and `CreateVisit.tsx` has no cycle with the bridge at all. As written, the comment tells the next reader the export was forced when it wasn't.

There is a real reason to prefer `bridgeLog` here, just not the documented one: it interpolates the error into the message string and so sidesteps the `JSON.stringify(new Error(…)) === "{}"` problem in `sendNativeLog`. Fixing that instead would let this stay private and keep the payload consistent with the `{message, data, level}` contract the sibling micro-apps use (`apps/menu-app/webapp/src/utils/microapp-bridge/index.ts:75-81`, `apps/people-app/microapp/src/components/microapp-bridge/index.ts:120-131`). `bridgeLog` omits `data` entirely, which I can't verify against the native handler from this repo.

Minor: the level union here is `"info" | "error" | "warn"` while `sendNativeLog`'s `LogLevel` (L101) includes `"debug"`. Two divergent level types in one file.

v15a1 (commented):
Post-merge review pass. `tsc --noEmit` is clean and the core fix is right: `decoded.name` was genuinely the missing claim, and `firstName = nameParts[0] || user.email` is load-bearing rather than cosmetic (with an empty name, `getOptionLabel(value)` returned `""` and MUI's value-sync effect wiped the input on the next unfocused render).

Eleven things the earlier rounds didn't cover, inline. The two that matter most:

1. The prepopulate effect is keyed on `[user]`, and `user` changes identity on every API request because the axios request interceptor refreshes the token on each one. The effect therefore re-runs constantly, and its `!whoTheyMeet` guard is exactly the state this PR's new feature creates. Clear the field, start typing, and your own name overwrites what you typed the moment the search fires. That largely defeats the second goal of the PR.

2. The commit that claims to drop PII from logs left two raw `${user.email}` interpolations on the native log channel.

The rest: an uncancelled async lookup that can display one host while submitting another, the pinned self option displacing a better API record, a MUI reason-string check that never matches on v7, `decoded.name` not trimmed at the boundary (which reopens the same class of bug this PR fixes), a dead catch block hiding a silent auth failure, and ~55 lines duplicated between the mount effect and `resetForm`.

Nothing here is a merge blocker in hindsight, but 1, 2 and the stale-write desync are worth a follow-up.

PR #309, opened 2026-07-17
[People App] Employees view follow-ups, office query fix, and rows-per-page persistence

shayanmalinda:
## Purpose

Follow-up improvements after #295, plus a production bug fix.

### Bug fix — GET /offices failed with `Unknown column 'working_locations'`
The `working_locations` JSON column was dropped from the `office` table, but `getOfficesQuery` still selected it, so the offices endpoint always errored. Removed the column from the query, the `Office` record, the webapp `Office` mapping, and the creation script (`people_app_creation.sql`). Also picks up patch-level `Dependencies.toml` bumps from a rebuild with Ballerina 2201.12.7.

### Employees view — active-filter chips and counting
- Extracted the styled multi-select filter chip into a reusable `RemovableFilterChip` component with a `dropFilterValue` helper.
- The baseline filter state (Active + Marked leaver, exclude future joiners) is no longer counted as an active filter, so the filter badge and "Clear all" only appear when the user has actually narrowed the search.

### Rows-per-page persistence
The rows-per-page value a user picks is stored in `localStorage` under one shared key (`people-app-rows-per-page`) and restored on mount, so it survives navigation/reloads and applies uniformly to the employees, my-team, and master-data tables. Values outside `PAGE_SIZE_OPTIONS` fall back to the default.

### Webapp auth header
The webapp now duplicates the ID token into the `x-jwt-assertion` header read by the backend `JwtInterceptor`, allowing the webapp to call the backend directly when no gateway injects the header.

## Verification
- `bal build` — backend compiles clean.
- `npx tsc --noEmit` and eslint over all touched webapp files — clean.

## Summary by CodeRabbit

- **New Features**
  - Data tables now remember your preferred rows-per-page setting between sessions.
  - Added reusable removable filter chips for employee search filters.
- **Improvements**
  - Refined active-filter indicators to avoid counting default employee status filters.
  - Simplified office information by removing working-location details.
- **Bug Fixes**
  - Improved authenticated API requests by including the required identity assertion.
  - Updated supporting package versions for improved reliability.

### DIFF
--- a/apps/people-app/webapp/src/view/employees/employeesView/employeesTable/EmployeesTable.tsx
+++ b/apps/people-app/webapp/src/view/employees/employeesView/employeesTable/EmployeesTable.tsx
@@ -37,10 +37,10 @@ export default function EmployeesTable() {
   const employeeState = useAppSelector((state) => state.employee);
   const navigate = useNavigate();
 
-  const [paginationModel, setPaginationModel] = useState<GridPaginationModel>({
+  const [paginationModel, setPaginationModel] = useState<GridPaginationModel>(() => ({
     page: 0,
     pageSize: getPersistedPageSize(),
-  });
+  }));
 
   const [sortModel, setSortModel] = useState<GridSortModel>([]);
 
--- a/apps/people-app/webapp/src/view/employees/employeesView/employeesTable/EmployeesTable.tsx
+++ b/apps/people-app/webapp/src/view/employees/employeesView/employeesTable/EmployeesTable.tsx
@@ -14,7 +14,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-import { DEFAULT_LIMIT_VALUE, PAGE_SIZE_OPTIONS } from "@config/constant";
+import { PAGE_SIZE_OPTIONS } from "@config/constant";
 import { alpha, Avatar, Box, Chip, Skeleton, Tooltip, useTheme } from "@mui/material";
 import { DataGrid, GridColDef, GridPaginationModel, GridRenderCellParams, GridSortItem, GridSortModel } from "@mui/x-data-grid";
 import {
@@ -29,7 +29,7 @@ import { State } from "@src/types/types";
 import { useEffect, useMemo, useState } from "react";
 import { useNavigate } from "react-router-dom";
 import { SearchForm } from "../searchForm/SearchForm";
-import { formatDate, getEmployeeStatusColor } from "@utils/utils";
+import { formatDate, getEmployeeStatusColor, getPersistedPageSize, persistPageSize } from "@utils/utils";
 
 export default function EmployeesTable() {
   const theme = useTheme();
@@ -39,7 +39,7 @@ export default function EmployeesTable() {
 
   const [paginationModel, setPaginationModel] = useState<GridPaginationModel>({
     page: 0,
-    pageSize: DEFAULT_LIMIT_VALUE,
+    pageSize: getPersistedPageSize(),
   });
 
   const [sortModel, setSortModel] = useState<GridSortModel>([]);
@@ -366,6 +366,7 @@ export default function EmployeesTable() {
           paginationModel={paginationModel}
           pageSizeOptions={PAGE_SIZE_OPTIONS}
           onPaginationModelChange={(model) => {
+            if (model.pageSize !== paginationModel.pageSize) persistPageSize(model.pageSize);
             setPaginationModel({
               page: model.page,
               pageSize: model.pageSize,
--- a/apps/people-app/webapp/src/view/employees/employeesView/searchForm/SearchForm.tsx
+++ b/apps/people-app/webapp/src/view/employees/employeesView/searchForm/SearchForm.tsx
@@ -72,6 +72,7 @@ import { State } from "@src/types/types";
 import { toSentenceCase, sortAndFormatOptions } from "@utils/utils";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { FilterChipSelect, FilterChipSelectProps } from "./FilterChipSelect";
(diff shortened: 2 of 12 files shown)


apps/people-app/webapp/src/utils/apiService.ts
```diff
@@ -90,6 +90,7 @@ export class APIService {
     APIService._instance.interceptors.request.use(
       (config) => {
         config.headers.set("Authorization", "Bearer " + APIService._idToken);
+        config.headers.set("X-JWT-Assertion", APIService._idToken);
```
shayanmalinda: **🟠 Question: Client-supplied `x-jwt-assertion` is trusted without signature validation**

The backend's `JwtInterceptor` only `jwt:decode`s this header (no signature check), so if the backend is ever reachable without a gateway that strips/overrides client-set `x-jwt-assertion`, any caller can forge a payload with admin groups. If this is only needed for local/direct development, consider gating it (e.g. behind a `window.config` flag) and confirming the deployed gateway overwrites the client copy.

---

apps/people-app/webapp/src/view/employees/employeesView/searchForm/SearchForm.tsx
```diff
@@ -238,6 +245,13 @@ export function SearchForm() {
       employeeStatuses,
       excludeFutureStartDate,
     } = filterPayload.filters;
+    // Baseline (Active + Marked leaver, exclude future joiners) is the default — don't count it.
+    const baselineStatuses = [EmployeeStatus.Active, EmployeeStatus.MarkedLeaver];
```
shayanmalinda: **🔵 Minor: Baseline-status check duplicated in two places**

This block is identical to the one in `hasAnyActiveFilters` (line 209), and both functions enumerate the same filters over the same input — the two copies will drift when the baseline changes. Extract a shared `isBaselineStatuses(employeeStatuses)` helper, or derive the boolean from the count (`hasActiveFilters = activeFilterCount > 0`) so the enumeration exists once.

> 📖 **Reference ([duplicate-code](https://refactoring.guru/smells/duplicate-code)):** "The same logic in more than one place; fixes/bugs must be applied N times and drift. Extract a shared function/module/factory."

---

apps/people-app/webapp/src/view/masterData/EntityTab.tsx
```diff
@@ -286,8 +287,9 @@ export default function EntityTab({
           loading={isLoading}
           localeText={{ noRowsLabel: `No ${entityLabel.toLowerCase()}s` }}
           initialState={{
-            pagination: { paginationModel: { pageSize: DEFAULT_LIMIT_VALUE } },
+            pagination: { paginationModel: { pageSize: getPersistedPageSize() } },
```
shayanmalinda, line 290: **🔵 Minor: Loading skeleton still sized by `DEFAULT_LIMIT_VALUE`**

The grid now starts at the persisted page size, but `SkeletonOverlay` (line 52) and the loading `minHeight` (line 322) still render `DEFAULT_LIMIT_VALUE` rows, so a user with 25 rows persisted sees a 10-row skeleton that jumps on load. Size both from `getPersistedPageSize()` (or the current model) instead.

---

apps/people-app/webapp/src/view/employees/employeesView/employeesTable/EmployeesTable.tsx
```diff
@@ -39,7 +39,7 @@ export default function EmployeesTable() {
 
   const [paginationModel, setPaginationModel] = useState<GridPaginationModel>({
     page: 0,
-    pageSize: DEFAULT_LIMIT_VALUE,
+    pageSize: getPersistedPageSize(),
```
shayanmalinda, line 42: **🔵 Minor: Non-lazy `useState` initializer re-reads localStorage every render**

`useState(getPersistedPageSize())` evaluates the localStorage read on every render and discards the result after mount; same in `MyTeamTable.tsx:48`. Use the lazy form: `useState<GridPaginationModel>(() => ({ page: 0, pageSize: getPersistedPageSize() }))`.

---

shayanmalinda (commented):
Reviewed the office-query fix, filter-chip refactor, rows-per-page persistence, and the new auth header. One security question and three minor cleanups.

apps/people-app/webapp/src/utils/apiService.ts
shayanmalinda: Removed — the `x-jwt-assertion` header is no longer sent from the webapp (commit dropped from this branch).

apps/people-app/webapp/src/view/employees/employeesView/searchForm/SearchForm.tsx
shayanmalinda: Fixed in 84eb30a9 — removed `hasAnyActiveFilters` entirely; the baseline check now lives in one module-level `isBaselineStatuses` helper and Clear-all visibility derives from the existing `active` flag.

apps/people-app/webapp/src/view/masterData/EntityTab.tsx
shayanmalinda: Fixed in 84eb30a9 — both the skeleton row count and the loading `minHeight` now use `getPersistedPageSize()`.

apps/people-app/webapp/src/view/employees/employeesView/employeesTable/EmployeesTable.tsx
shayanmalinda: Fixed in 84eb30a9 — both tables now use the lazy initializer form.

shayanmalinda:
Closing to keep the open changeset minimal — the urgent office-fetch fix is raised separately as #310. The employees-view follow-ups and rows-per-page persistence on this branch will be re-raised later.

PR #311, opened 2026-07-21
[People App] Prevent duplicate car park bookings and double charges

shayanmalinda:
## Problem

A user who hit a **"payment failed"** error in the Car Park Booking flow and re-submitted ended up with **two confirmed bookings, two on-chain transactions, and two Google Sheet rows** for the same day (charged 0.04 O2C twice for one physical slot). Reported by a user with a screenshot of two `CONFIRMED` bookings.

## Root cause

Two defects combined:

1. **Trigger — false "payment failed" error.** The wallet payment is confirmed by **two independent drivers sharing the same bridge keys**: the inline poll in `ParkingBookingSummaryPage.handleConfirmAndPay` and the globally-mounted `ParkingWalletReturnResume` (`App.tsx`). On return from the wallet they race — the winner confirms the reservation and clears the bridge keys, so the loser either times out (`"Payment timed out"`) or gets `"already CONFIRMED"` and shows **Payment Unsuccessful** on a booking that actually succeeded and was already charged. That false error is what made the user retry.
2. **Root cause — no per-user guard.** Uniqueness was enforced **per slot only** (`uk_active_slot_booking_date`), never per user. The retry landed on a **different slot**, so `createReservation` succeeded, the wallet transferred a **second** time, and a second booking was confirmed.

## Changes

**Backend — authoritative guarantee**
- New functional unique index `uk_active_employee_booking_date`: at most one `PENDING`/`CONFIRMED` reservation per employee per booking date (migration `v1.0.21` + `people_app_creation.sql`), mirroring the existing slot index (EXPIRED excluded).
- New DB layer: `getActiveParkingReservationForEmployeeDate`, `expireStalePendingParkingReservationsForEmployeeDate`, `ActiveParkingReservationRow`, and a `DuplicateActiveReservationError` distinct error; `addParkingReservation` maps MySQL `1062` to it (structured, mirroring `utils.bal:getErrorCode`).
- `POST /parkings/reservations`: expires the caller's stale pendings, **reuses** a same-slot `PENDING` (idempotent retry — no second charge), returns **400** if the caller already holds a booking for the date, and maps the unique-index race to a clean 400. A retry on a new slot now fails **before** the wallet opens.

**Frontend — kills the false error and re-charge (`ParkingBookingSummaryPage.handleConfirmAndPay`)**
- Reuses an existing `reservationId` instead of blindly re-creating; before charging, finishes without paying again when the reservation is already `CONFIRMED` or a prior wallet payment succeeded but wasn't confirmed.
- The `catch` now **verifies server state before showing failure**, so a booking that actually went through no longer surfaces a spurious error.

## Verification

- **Microapp:** `tsc -b`, `eslint`, and `vite build` all pass.
- **Backend:** could **not** be compiled locally — the installed `bal` is `2201.7.5`, which crashes (`NoSuchElementException` in `RegExpParser.parseInterpolation`) on the pre-existing `re \`\${EMAIL_PATTERN}\`` in `types.bal` while scanning imports; this reproduces on a pristine `main`, so it is an environment issue (project targets `2201.12.7`), not this change. Please confirm via CI. The Ballerina edits follow existing patterns exactly.

## Not in scope
- The 0.04 O2C refund to the affected user is a manual wallet ops action.

## Summary by CodeRabbit

* **New Features**
  * Employees can now have only one active parking reservation per day (pending/confirmed).
  * Repeating a request can reuse your existing pending reservation when it matches the same day and slot.
  * Stale pending reservations are automatically expired before new bookings.
  * Slot selection now shows an “already booked today” notice with “View My Bookings” and disables starting another booking while checking.

* **Bug Fixes**
  * Improved confirm-and-pay recovery to prevent duplicate charges and finalize confirmation safely after interruptions.
  * Clearer handling when an active reservation conflict occurs (refresh and retry).

### DIFF
--- a/apps/people-app/backend/resources/people_app_table_update_v1.0.21.sql
+++ b/apps/people-app/backend/resources/people_app_table_update_v1.0.21.sql
@@ -9,20 +9,27 @@
 -- paid reservation and charged the wallet twice.
 --
 -- This functional unique index mirrors uk_active_slot_booking_date but is keyed
--- on the employee instead of the slot. EXPIRED rows map to NULL so they are
--- ignored by the index, allowing a fresh booking once a prior attempt expires.
+-- on the employee instead of the slot. Non-active rows (EXPIRED, DUPLICATE) map
+-- to NULL so they are ignored by the index, allowing a fresh booking once a
+-- prior attempt is no longer active.
 --
 -- Pre-cleanup: because this bug allowed multiple active reservations per
 -- employee/date, production may already hold such duplicates (the reported
 -- incident is one). CREATE UNIQUE INDEX would fail (error 1062) on them, so
--- first expire all-but-one active row per (employee_email, booking_date),
--- keeping the CONFIRMED row when present, otherwise the most recent id. The
--- derived table is materialized, so updating parking_reservation while reading
--- it in the subquery does not hit MySQL error 1093.
+-- first collapse all-but-one active row per (employee_email, booking_date) to a
+-- new DUPLICATE status, keeping the CONFIRMED row when present, otherwise the
+-- most recent id. DUPLICATE (rather than EXPIRED) preserves the fact that these
+-- were genuine, often paid, bookings for audit/refund purposes. The derived
+-- table is materialized, so updating parking_reservation while reading it in the
+-- subquery does not hit MySQL error 1093.
 -- =====================================================================
 
+-- Add the DUPLICATE status before it is assigned below.
+ALTER TABLE `parking_reservation`
+    MODIFY COLUMN `status` ENUM ('PENDING', 'CONFIRMED', 'EXPIRED', 'DUPLICATE') NOT NULL DEFAULT 'PENDING';
+
 UPDATE `parking_reservation`
-SET `status` = 'EXPIRED'
+SET `status` = 'DUPLICATE'
 WHERE `status` IN ('PENDING', 'CONFIRMED')
   AND `id` NOT IN (
     SELECT `keep_id` FROM (
--- a/apps/people-app/backend/resources/people_app_table_update_v1.0.21.sql
+++ b/apps/people-app/backend/resources/people_app_table_update_v1.0.21.sql
@@ -11,7 +11,31 @@
 -- This functional unique index mirrors uk_active_slot_booking_date but is keyed
 -- on the employee instead of the slot. EXPIRED rows map to NULL so they are
 -- ignored by the index, allowing a fresh booking once a prior attempt expires.
+--
+-- Pre-cleanup: because this bug allowed multiple active reservations per
+-- employee/date, production may already hold such duplicates (the reported
+-- incident is one). CREATE UNIQUE INDEX would fail (error 1062) on them, so
+-- first expire all-but-one active row per (employee_email, booking_date),
+-- keeping the CONFIRMED row when present, otherwise the most recent id. The
+-- derived table is materialized, so updating parking_reservation while reading
+-- it in the subquery does not hit MySQL error 1093.
 -- =====================================================================
+
+UPDATE `parking_reservation`
+SET `status` = 'EXPIRED'
+WHERE `status` IN ('PENDING', 'CONFIRMED')
+  AND `id` NOT IN (
+    SELECT `keep_id` FROM (
+        SELECT `id` AS `keep_id`,
+               ROW_NUMBER() OVER (
+                   PARTITION BY `employee_email`, `booking_date`
+                   ORDER BY (`status` = 'CONFIRMED') DESC, `id` DESC
+               ) AS `rn`
+        FROM `parking_reservation`
+        WHERE `status` IN ('PENDING', 'CONFIRMED')
+    ) `ranked`
+    WHERE `rn` = 1
+  );
 
 CREATE UNIQUE INDEX `uk_active_employee_booking_date` ON `parking_reservation` ((
     CASE
--- a/apps/people-app/backend/resources/people_app_table_update_v1.0.21.sql
+++ b/apps/people-app/backend/resources/people_app_table_update_v1.0.21.sql
@@ -1,0 +1,22 @@
+-- =====================================================================
+-- people_app_table_update_v1.0.21.sql
+-- Fix: prevent a single employee from holding more than one active
+--      (PENDING or CONFIRMED) parking reservation per booking date.
+--
+-- Until now uniqueness was enforced per slot only (uk_active_slot_booking_date,
+-- v1.0.11). Nothing stopped one employee from booking multiple slots on the
+-- same day, so a retry after a spurious "booking failed" error created a second
+-- paid reservation and charged the wallet twice.
+--
+-- This functional unique index mirrors uk_active_slot_booking_date but is keyed
+-- on the employee instead of the slot. EXPIRED rows map to NULL so they are
+-- ignored by the index, allowing a fresh booking once a prior attempt expires.
+-- =====================================================================
+
+CREATE UNIQUE INDEX `uk_active_employee_booking_date` ON `parking_reservation` ((
+    CASE
+        WHEN status IN ('PENDING', 'CONFIRMED')
+            THEN CONCAT(employee_email, '|', booking_date)
+        ELSE NULL
+    END
+));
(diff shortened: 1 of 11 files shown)


apps/people-app/backend/resources/people_app_table_update_v1.0.21.sql
```diff
@@ -0,0 +1,22 @@
...
+-- v1.0.11). Nothing stopped one employee from booking multiple slots on the
+-- same day, so a retry after a spurious "booking failed" error created a second
+-- paid reservation and charged the wallet twice.
+--
+-- This functional unique index mirrors uk_active_slot_booking_date but is keyed
+-- on the employee instead of the slot. EXPIRED rows map to NULL so they are
+-- ignored by the index, allowing a fresh booking once a prior attempt expires.
+-- =====================================================================
+
+CREATE UNIQUE INDEX `uk_active_employee_booking_date` ON `parking_reservation` ((
```
shayanmalinda, line 47: **🔴 High: Index creation will fail on pre-existing duplicate active reservations**

`CREATE UNIQUE INDEX` aborts with `ERROR 1062 Duplicate entry` if the table already holds two `PENDING`/`CONFIRMED` rows for the same `(employee_email, booking_date)` — which is exactly what this bug produced in production (the reported incident is one such pair), so the migration will fail on deploy. Add a pre-cleanup before the `CREATE` that expires all-but-one active row per `(employee_email, booking_date)`, preferring to keep the `CONFIRMED` one.

---

apps/people-app/backend/service.bal
```diff
@@ -1848,6 +1848,48 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+                wso2_coin:pendingReservationExpiryMinutes);
+        if activeReservation is error {
+            log:printError("Error checking existing reservations for employee", activeReservation);
+            return <http:InternalServerError>{
+                body: {message: "Error occurred while validating booking eligibility."}
+            };
+        }
+        if activeReservation is database:ActiveParkingReservationRow {
+            // Same-slot PENDING: reuse it so a retry does not create (or pay for) a second booking.
+            if activeReservation.status == database:PENDING && activeReservation.slotId == body.slotId {
```
shayanmalinda, line 1874: **🟡 Medium: Same-slot reuse ignores the vehicle selected on retry**

When an existing same-slot `PENDING` reservation is reused, `body.vehicleId` is discarded, so if the user changes vehicle before retrying, the booking and its Google Sheet row keep the original vehicle. Either update the reservation's `vehicle_id` on reuse, or only reuse when the existing reservation's vehicle matches `body.vehicleId`.

---

shayanmalinda (commented):
Automated review: 2 findings (1 high — migration will fail on the duplicate rows this bug already created; 1 medium — reuse ignores the retry vehicle).

apps/people-app/backend/resources/people_app_table_update_v1.0.21.sql
shayanmalinda: Added a pre-cleanup UPDATE before the `CREATE UNIQUE INDEX` that expires all-but-one active row per `(employee_email, booking_date)` (keeping the CONFIRMED one, else most recent `id`), via a materialized derived table to avoid MySQL 1093. (362fdf80)

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

apps/people-app/backend/service.bal
shayanmalinda: Fixed: added `vehicleId` to `ActiveParkingReservationRow` and, on same-slot reuse, the reservation's `vehicle_id` is now updated to `body.vehicleId` when it differs (new `updateParkingReservationVehicle`). (362fdf80)

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

apps/people-app/microapp/src/pages/ParkingBookingSummaryPage.tsx
```diff
@@ -246,15 +247,65 @@ function ParkingBookingSummaryPage() {
...
+        const reservation = await createReservation();
+        reservationId = reservation.reservationId;
+        coinsAmount = reservation.coinsAmount;
+
+        setParkingPaymentContextState({
+          ...paymentContext,
+          reservationId,
+          coinsAmount,
+        });
+      }
```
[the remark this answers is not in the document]
shayanmalinda: Good catch — valid. Restructured the recovery branch: only a PENDING reservation is reused (and confirmed with a prior wallet txHash when present); CONFIRMED finishes without charging; any other status (e.g. EXPIRED after the pending window lapses) now discards the stale `reservationId` and creates a fresh PENDING row before Stage 2, so the wallet is never charged against a non-confirmable reservation. (9c8819ee)

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

apps/people-app/microapp/src/pages/ParkingSlotSelectionPage.tsx
```diff
@@ -232,13 +239,49 @@ function ParkingSlotSelectionPage() {
...
+        setExistingBooking(null);
+        setCheckingExistingBooking(false);
+      },
+      () => {},
+    );
+    return () => {
+      cancelled = true;
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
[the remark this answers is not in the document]
shayanmalinda: Keyed the effect on `todayBookingDate` so the guard refreshes on a date rollover rather than capturing the initial value. Kept `executeWithTokenHandling` rather than converting to `useQuery` — every fetch on this page (floors, slots, vehicles, config) uses that helper, so a lone `useQuery` would diverge from the file's established pattern; happy to migrate the whole page separately if desired. (4c13c532)

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

apps/people-app/microapp/src/pages/ParkingSlotSelectionPage.tsx
```diff
@@ -477,19 +520,37 @@ function ParkingSlotSelectionPage() {
...
+                    </div>
+                    <button
+                      type="button"
+                      className="mt-3 w-full p-[0.75rem] text-[14px] font-semibold rounded-[0.7rem] border border-[#E5E5E5] bg-white text-[#1F2A44]"
+                      onClick={() => navigate("/services/parking/bookings")}
+                    >
+                      View My Bookings
+                    </button>
+                  </>
+                )}
```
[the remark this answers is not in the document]
shayanmalinda: Fixed — the notice and View My Bookings action now render independently whenever an active booking exists, before slot selection and regardless of the booking window; the window-closed warning and slot panel are gated on `!existingBooking` so they don't compete. (4c13c532)

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

apps/people-app/microapp/src/pages/ParkingSlotSelectionPage.tsx
```diff
@@ -232,13 +239,49 @@ function ParkingSlotSelectionPage() {
...
+      handleRequest,
+      handleRequestWithNewToken,
+      serviceUrls.fetchParkingReservations(todayBookingDate, todayBookingDate),
+      "GET",
+      null,
+      (data) => {
+        if (cancelled) return;
+        const list = (data as ParkingReservationDetails[]) ?? [];
+        const active = list.find(
+          (r) => r.status === "PENDING" || r.status === "CONFIRMED",
```
shayanmalinda: **🟡 Medium: Stale PENDING falsely blocks re-booking (frontend vs backend "active" mismatch)**

The backend guard (`getActiveParkingReservationForEmployeeDate`) counts a PENDING as active only within `pendingExpiryMinutes`, and expires stale ones on the next create — but this filter treats *any* PENDING as blocking, so a user with an abandoned pending older than the window is stuck ("You already have a booking today") even though the slot shows available and the backend would let them book. Mirror the backend's freshness rule (block on `CONFIRMED` always, `PENDING` only when fresh) or defer the pending case to the backend.

---

shayanmalinda (commented):
Re-review of the full PR: 1 finding — the frontend booking guard treats any PENDING as blocking, but the backend only counts fresh pendings, so a stale abandoned pending falsely blocks re-booking.

apps/people-app/microapp/src/pages/ParkingSlotSelectionPage.tsx
shayanmalinda: Fixed — the guard now blocks only on a CONFIRMED reservation (`list.find((r) => r.status === 'CONFIRMED')`). The PENDING case is deferred to the backend, which reuses a same-slot pending and expires stale ones, so an abandoned pending no longer falsely blocks re-booking. (author decision: keep it minimal and timezone-risk-free) (4d14bb70)

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

Rashmika998 (approved):

PR #312, opened 2026-07-23
[People App] Fix duplicate employee entries returned to UI

shayanmalinda:
## Purpose

The employees list was returning **duplicate employee rows** to the UI.

## Root cause

Duplicate `employee` records sharing the same `work_email` (surfaced by the ex-employee migration backfill) meant that the manager JOIN in `getEmployeesQuery` — `LEFT JOIN employee mgr ON LOWER(e.manager_email) = LOWER(mgr.work_email)` — matched more than one manager row per employee, multiplying each employee into several rows in the result set.

## Fix

`backend/modules/database/db_queries.bal`:

- **`getEmployeesQuery`** — replace the direct manager JOIN with a derived table that groups by `LOWER(work_email)` and picks the latest name by `id`, so every employee matches exactly one manager row.
- **`getEmployeeInfoQuery`** — replace the manager JOIN with a correlated subquery returning the latest manager name by `id` (`ORDER BY id DESC LIMIT 1`).
- **`getManagersQuery`** — switch `SELECT DISTINCT` to `GROUP BY m.work_email`, picking the latest `employee_id` by `id`, to avoid duplicate managers.

## Scope

Single-file, query-only change. No schema, API contract, or frontend changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved manager name accuracy and consistency in employee records.
  * Ensured the most recent manager information is selected when duplicate or historical records exist.
  * Added safer handling for missing manager names, displaying a blank value instead of an incorrect result.

### DIFF
--- a/apps/people-app/backend/modules/database/db_queries.bal
+++ b/apps/people-app/backend/modules/database/db_queries.bal
@@ -386,10 +386,12 @@ isolated function getEmployeesQuery(EmployeeSearchPayload payload, string? leadE
 # + return - Parameterized query for fetching distinct managers
 isolated function getManagersQuery() returns sql:ParameterizedQuery =>
     `SELECT
-        SUBSTRING_INDEX(GROUP_CONCAT(m.employee_id ORDER BY m.id DESC SEPARATOR '||'), '||', 1) AS employee_id,
+        SUBSTRING_INDEX(
+            GROUP_CONCAT(m.employee_id ORDER BY (m.employee_status = 'Active') DESC, m.id DESC SEPARATOR '||'), '||', 1
+        ) AS employee_id,
         m.work_email
     FROM employee e
-    JOIN employee m ON e.manager_email = m.work_email
+    JOIN employee m ON LOWER(e.manager_email) = LOWER(m.work_email)
     WHERE e.manager_email IS NOT NULL AND e.manager_email <> ''
     GROUP BY m.work_email;`;
 
--- a/apps/people-app/backend/modules/database/db_queries.bal
+++ b/apps/people-app/backend/modules/database/db_queries.bal
@@ -389,11 +389,13 @@ isolated function getManagersQuery() returns sql:ParameterizedQuery =>
         SUBSTRING_INDEX(
             GROUP_CONCAT(m.employee_id ORDER BY (m.employee_status = 'Active') DESC, m.id DESC SEPARATOR '||'), '||', 1
         ) AS employee_id,
-        m.work_email
+        SUBSTRING_INDEX(
+            GROUP_CONCAT(m.work_email ORDER BY (m.employee_status = 'Active') DESC, m.id DESC SEPARATOR '||'), '||', 1
+        ) AS work_email
     FROM employee e
     JOIN employee m ON LOWER(e.manager_email) = LOWER(m.work_email)
     WHERE e.manager_email IS NOT NULL AND e.manager_email <> ''
-    GROUP BY m.work_email;`;
+    GROUP BY LOWER(m.work_email);`;
 
 # Check if a target employee is a direct or additional subordinate of a lead.
 #
--- a/apps/people-app/backend/modules/database/db_queries.bal
+++ b/apps/people-app/backend/modules/database/db_queries.bal
@@ -88,7 +88,13 @@ isolated function getEmployeeInfoQuery(string employeeId) returns sql:Parameteri
         e.work_location AS workLocation,
         e.start_date AS startDate,
         e.manager_email AS managerEmail,
-        COALESCE(CONCAT(mgr.first_name, ' ', mgr.last_name), '') AS managerName,
+        COALESCE((
+            SELECT CONCAT(m.first_name, ' ', m.last_name)
+            FROM employee m
+            WHERE LOWER(m.work_email) = LOWER(e.manager_email)
+            ORDER BY m.id DESC
+            LIMIT 1
+        ), '') AS managerName,
         COALESCE(eam.additionalManagerEmails, '') AS additionalManagerEmails,
         pi.gender AS gender,
         (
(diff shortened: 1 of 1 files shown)


apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -372,12 +385,13 @@ isolated function getEmployeesQuery(EmployeeSearchPayload payload, string? leadE
 #
 # + return - Parameterized query for fetching distinct managers
 isolated function getManagersQuery() returns sql:ParameterizedQuery =>
-    `SELECT DISTINCT 
-        m.employee_id, 
+    `SELECT
+        SUBSTRING_INDEX(GROUP_CONCAT(m.employee_id ORDER BY m.id DESC SEPARATOR '||'), '||', 1) AS employee_id,
```
shayanmalinda: **🟡 Question: `getManagersQuery` may return an ex-employee's `employee_id`**

Among duplicate records sharing a `work_email`, the `ORDER BY m.id DESC` tiebreak picks the highest-id row, and there is no `employee_status` filter — if the ex-employee backfill inserts historical rows with higher ids, this returns the ex-employee's `employee_id` rather than the active manager's. Should this prefer the active record (and match `work_email` case-insensitively with `LOWER()`, like the sibling queries in this PR)?

---

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -253,7 +258,15 @@ isolated function getEmployeesQuery(EmployeeSearchPayload payload, string? leadE
             LEFT JOIN sub_team st ON st.id = e.sub_team_id
             LEFT JOIN unit u ON u.id = e.unit_id
             LEFT JOIN house h ON h.id = e.house_id
-            LEFT JOIN employee mgr ON LOWER(e.manager_email) = LOWER(mgr.work_email)
+            LEFT JOIN (
+                SELECT
+                    LOWER(work_email) AS managerEmail,
+                    SUBSTRING_INDEX(
+                        GROUP_CONCAT(CONCAT(first_name, ' ', last_name) ORDER BY id DESC SEPARATOR '||'), '||', 1
```
shayanmalinda, line 265: **🔵 Minor: `GROUP_CONCAT`+`SUBSTRING_INDEX` to pick the latest name**

`getEmployeeInfoQuery` resolves the identical "latest manager name by email" with a clean correlated subquery (`ORDER BY id DESC LIMIT 1`), while this derived table emulates top-1 via string concatenation — harder to read, and it truncates if a name ever contains `||` or the group exceeds `group_concat_max_len`. Since duplicate rows share the same manager the names are identical anyway, so `MAX(...)`/`ANY_VALUE(...)` (or a window function) would express the intent more directly.

---

shayanmalinda (commented):
Reviewed the manager-row dedup fix. The core change is sound — the derived table yields exactly one manager row per employee, eliminating the JOIN fan-out that duplicated rows. Two questions on the dedup tiebreak and query consistency below.

apps/people-app/backend/modules/database/db_queries.bal
shayanmalinda: Fixed in b2082ea0. `getManagersQuery` now joins **and** groups case-insensitively (`LOWER(...)`), and the dedup tiebreak prefers the `Active` record (falling back to latest `id`) for both `employee_id` and `work_email` — so an ex-employee's id is no longer returned for a manager.

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -372,12 +385,13 @@ isolated function getEmployeesQuery(EmployeeSearchPayload payload, string? leadE
...
-    `SELECT DISTINCT 
-        m.employee_id, 
+    `SELECT
+        SUBSTRING_INDEX(GROUP_CONCAT(m.employee_id ORDER BY m.id DESC SEPARATOR '||'), '||', 1) AS employee_id,
         m.work_email
     FROM employee e
     JOIN employee m ON e.manager_email = m.work_email
-    WHERE e.manager_email IS NOT NULL AND e.manager_email <> '';`;
+    WHERE e.manager_email IS NOT NULL AND e.manager_email <> ''
+    GROUP BY m.work_email;`;
```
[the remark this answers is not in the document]
shayanmalinda: Applied in b2082ea0 — joined and grouped on `LOWER(...)` and selected the chosen rows `employee_id` and `work_email` via the aggregation. One deviation from the suggested diff: kept `ORDER BY (m.employee_status = 'Active') DESC, m.id DESC` as the tiebreak so the **Active** record wins over a stale ex-employee duplicate, not merely the latest `id`.

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

apps/people-app/backend/modules/database/db_queries.bal
shayanmalinda: Keeping the `GROUP_CONCAT(... ORDER BY id DESC)` approach here. The `ORDER BY id DESC` deliberately selects the *most recent* records name — `MAX()`/`ANY_VALUE()` would lose that latest-wins semantics (`MAX` returns the alphabetically-last name, `ANY_VALUE` an arbitrary one). This greatest-n-per-group idiom is already the established pattern in this file (`additionalManagerEmails` at lines 146/235/475), and the fragility is negligible: person names dont contain `||`, and each managers group is only 1-3 rows — far below `group_concat_max_len`.

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

Rashmika998 (approved):

kasunsiyambalapitiya (approved):

dilshanfardil (approved):

PR #316, opened 2026-07-30
[People Ops Suite][BE] Support people backend to have external user groups

Rashmika998:
## Description
- Support people backend to have external user groups for vehicle saving purposes.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring roles for external users.
  * Improved authorization handling for nested and multiple role assignments.

* **Bug Fixes**
  * Authorization now more reliably matches users’ group memberships against their permitted roles.

### DIFF
--- a/apps/people-app/backend/Config.toml.local
+++ b/apps/people-app/backend/Config.toml.local
@@ -2,6 +2,7 @@
 EMPLOYEE_ROLE = ""
 ADMIN_ROLE = ""
 SERVICE_DESK_ROLE = ""
+EXTERNAL_USER_ROLES = [];
 
 [people.database.dbConfig]
     host = ""
--- a/apps/people-app/backend/modules/authorization/authorization.bal
+++ b/apps/people-app/backend/modules/authorization/authorization.bal
@@ -52,24 +52,39 @@ public isolated service class JwtInterceptor {
             return <http:InternalServerError>{body: {message: errorMsg}};
         }
 
-        // Flatten the authorized roles into a single array for easier comparison.
-        string[] allAuthorizedRoles = [];
-        foreach string|string[] role in authorizedRoles.toArray() {
-            if role is string[] {
-                foreach string r in role {
-                    allAuthorizedRoles.push(r);
-                }
-            } else {
-                allAuthorizedRoles.push(role);
+        // Internal roles are allowed on any resource (further scoped by per-resource checks in service.bal).
+        string[] internalRoles = [authorizedRoles.EMPLOYEE_ROLE, authorizedRoles.ADMIN_ROLE,
+            authorizedRoles.SERVICE_DESK_ROLE];
+        boolean hasInternalRole = false;
+        foreach string userGroup in userInfo.groups {
+            if internalRoles.indexOf(userGroup) !is () {
+                hasInternalRole = true;
+                break;
             }
         }
+        if hasInternalRole {
+            ctx.set(HEADER_USER_INFO, userInfo);
+            return ctx.next();
+        }
 
-        // For regular user tokens, we check if they have the required roles to access the resource.
+        // External users are only allowed to access the vehicles resource of their own record.
+        boolean hasExternalRole = false;
         foreach string userGroup in userInfo.groups {
-            if allAuthorizedRoles.indexOf(userGroup) !is () {
+            if authorizedRoles.EXTERNAL_USER_ROLES.indexOf(userGroup) !is () {
+                hasExternalRole = true;
+                break;
+            }
+        }
+        if hasExternalRole {
+            if path.length() == 3 && path[0] == "employees" && path[2] == "vehicles"
+                    && (req.method == http:HTTP_GET || req.method == http:HTTP_POST) {
                 ctx.set(HEADER_USER_INFO, userInfo);
                 return ctx.next();
             }
+
+            log:printWarn(string `${userInfo.email} (external user) attempted to access a restricted resource: `
+                    + string `${req.method} ${path.toBalString()}`);
+            return <http:Forbidden>{body: {message: "Insufficient privileges!"}};
         }
 
         log:printError(
--- a/apps/people-app/backend/modules/authorization/authorization.bal
+++ b/apps/people-app/backend/modules/authorization/authorization.bal
@@ -60,7 +60,7 @@ public isolated service class JwtInterceptor {
                     allAuthorizedRoles.push(r);
                 }
             } else {
-                allAuthorizedRoles.push(<string>role);
+                allAuthorizedRoles.push(role);
             }
         }
 
--- a/apps/people-app/backend/modules/authorization/authorization.bal
+++ b/apps/people-app/backend/modules/authorization/authorization.bal
@@ -52,8 +52,21 @@ public isolated service class JwtInterceptor {
             return <http:InternalServerError>{body: {message: errorMsg}};
         }
 
-        foreach anydata role in authorizedRoles.toArray() {
-            if userInfo.groups.some(r => r === role) {
+        // Flatten the authorized roles into a single array for easier comparison.
+        string[] allAuthorizedRoles = [];
+        foreach string|string[] role in authorizedRoles.toArray() {
+            if role is string[] {
+                foreach string r in role {
+                    allAuthorizedRoles.push(r);
+                }
+            } else {
+                allAuthorizedRoles.push(<string>role);
+            }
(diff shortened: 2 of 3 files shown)


shayanmalinda (commented):
Reviewed the external-user-groups change. Main question inline: EXTERNAL_USER_ROLES opens the interceptor gate for the whole API but is never checked per-resource (unlike visitor-app's pattern); plus three minor notes (Config.toml.local template, a redundant cast, per-request flattening).

apps/people-app/backend/modules/authorization/authorization.bal
```diff
@@ -52,8 +52,21 @@ public isolated service class JwtInterceptor {
...
+                    allAuthorizedRoles.push(r);
+                }
+            } else {
+                allAuthorizedRoles.push(<string>role);
+            }
+        }
+
+        // For regular user tokens, we check if they have the required roles to access the resource.
+        foreach string userGroup in userInfo.groups {
+            if allAuthorizedRoles.indexOf(userGroup) !is () {
```
shayanmalinda: **🔴 High: External users pass the gate for every endpoint — `EXTERNAL_USER_ROLES` is never checked per-resource**

The interceptor admits any user holding any configured role, and nothing in `service.bal` consumes `EXTERNAL_USER_ROLES`, so an external (vehicle) user can reach every endpoint that has no per-resource role check — `parkings/*`, `organization-structure`, `companies`, `offices`, `employment-types`, `user-info` — not just the vehicle ones. visitor-app pairs this same gate with per-resource scoping (`checkPermissions([authorization:authorizedRoles.EXTERNAL_USER_ROLE], ...)`, e.g. `apps/visitor-app/backend/service.bal:77`) — is the same restriction planned here for the endpoints external users may actually use?

---

apps/people-app/backend/modules/authorization/types.bal
```diff
@@ -30,4 +30,6 @@ public type AppRoles record {|
     string ADMIN_ROLE;
     # Role for the service desk
     string SERVICE_DESK_ROLE;
+    # Roles for the external users(for vehicle saving purpose)
+    string[] EXTERNAL_USER_ROLES = [];
```
shayanmalinda, line 34: **🟡 Minor: `Config.toml.local` not updated with the new field**

`EXTERNAL_USER_ROLES` was added to `AppRoles` but `apps/people-app/backend/Config.toml.local` still lists only the three existing roles, so developers setting up locally won't know it exists. Add a placeholder, e.g. `EXTERNAL_USER_ROLES = [""]` — visitor-app's template already lists its `EXTERNAL_USER_ROLE` the same way.

> 📖 **Reference (Ballerina reference — Configuration):** "When adding new configurable sections to `Config.toml`, always update `Config.toml.local` (the template) with placeholder values — this is the reference for other developers setting up locally"

---

apps/people-app/backend/modules/authorization/authorization.bal
```diff
@@ -52,8 +52,21 @@ public isolated service class JwtInterceptor {
...
-            if userInfo.groups.some(r => r === role) {
+        // Flatten the authorized roles into a single array for easier comparison.
+        string[] allAuthorizedRoles = [];
+        foreach string|string[] role in authorizedRoles.toArray() {
+            if role is string[] {
+                foreach string r in role {
+                    allAuthorizedRoles.push(r);
+                }
+            } else {
+                allAuthorizedRoles.push(<string>role);
```
shayanmalinda: **🟡 Minor: Redundant `<string>` cast**

In the `else` branch `role` is already narrowed to `string` by the `if role is string[]` check, so the cast is unnecessary — `allAuthorizedRoles.push(role);`.

> 📖 **Reference ([avoid_unnecessary_type_casts](https://learn-ballerina.github.io/best_practices/avoid_unnecessary_type_casts.html)):** "Avoid unnecessary type casts"

---

apps/people-app/backend/modules/authorization/authorization.bal
```diff
@@ -52,8 +52,21 @@ public isolated service class JwtInterceptor {
             return <http:InternalServerError>{body: {message: errorMsg}};
         }
 
-        foreach anydata role in authorizedRoles.toArray() {
-            if userInfo.groups.some(r => r === role) {
+        // Flatten the authorized roles into a single array for easier comparison.
+        string[] allAuthorizedRoles = [];
```
shayanmalinda: **🟡 Minor: Roles are re-flattened on every request**

`authorizedRoles` is a configurable and can't change after startup, so rebuilding `allAuthorizedRoles` inside the interceptor repeats pure setup work on every request's auth path. Compute it once at module level instead, e.g. `final readonly & string[] allAuthorizedRoles = flattenRoles(authorizedRoles);` (same applies to the visitor-app original this was ported from).

---

apps/people-app/backend/modules/authorization/authorization.bal
Rashmika998: did it similar to visitor app. I'll fix

apps/people-app/backend/modules/authorization/authorization.bal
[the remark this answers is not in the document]
Rashmika998: As discussed offline i believe this is fine. @shayanmalinda @sacheeramesh

Rashmika998: [ditto](https://github.com/wso2-open-operations/people-ops-suite/pull/316#discussion_r3679962511)

sacheeramesh:
This would allow external users to access any APIs that aren't protected by the custom authorization logic have to fix them.
Shall we apply the same custom authorization to the vehicle endpoint so users can only retrieve and create their own vehicle records from this external role?

sacheeramesh:
Or we can limit access to a specific resource path from the authorization module.

Rashmika998:
> This would allow external users to access any APIs that aren't protected by the custom authorization logic have to fix them. Shall we apply the same custom authorization to the vehicle endpoint so users can only retrieve and create their own vehicle records from this external role?

Fixed

apps/people-app/backend/modules/authorization/authorization.bal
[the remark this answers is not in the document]
Rashmika998: fixed

Rashmika998: fixed

shayanmalinda (commented):
Post-fix verification: all 5 review threads genuinely addressed in the head code; two residual items from the fix commits below (invalid TOML in Config.toml.local, and a groups-claim edge for external-store tokens).

apps/people-app/backend/Config.toml.local
```diff
@@ -2,6 +2,7 @@
 EMPLOYEE_ROLE = ""
 ADMIN_ROLE = ""
 SERVICE_DESK_ROLE = ""
+EXTERNAL_USER_ROLES = [];
```
shayanmalinda, line 5: **🟠 Medium: Invalid TOML — trailing semicolon**

TOML has no statement terminator, so `EXTERNAL_USER_ROLES = [];` fails to parse and anyone copying this template gets a config parse error on `bal run`. Drop the semicolon and use a placeholder entry: `EXTERNAL_USER_ROLES = [""]`, matching visitor-app's template.

---

apps/people-app/backend/modules/authorization/authorization.bal
```diff
@@ -52,11 +52,39 @@ public isolated service class JwtInterceptor {
...
+        }
+        if hasInternalRole {
+            ctx.set(HEADER_USER_INFO, userInfo);
+            return ctx.next();
+        }
+
+        // External users are only allowed to access the vehicles resource of their own record.
+        boolean hasExternalRole = false;
+        foreach string userGroup in userInfo.groups {
+            if authorizedRoles.EXTERNAL_USER_ROLES.indexOf(userGroup) !is () {
```
shayanmalinda, line 73: **🟡 Minor: External-store tokens without a `groups` claim get a 500, not a 403**

people-app's `CustomJwtPayload` declares `string[] groups;` with no default, so a token carrying no `groups` claim fails `cloneWithType` earlier in this interceptor and returns 500 "Malformed Invoker info object!". visitor-app defaults the field for exactly this audience (`string[] groups = []; // External user store users may not have groups claim in the token`) — worth the same default here now that external users are expected callers.

---

shayanmalinda (approved):

PR #317, opened 2026-07-30
People App - feat(employee): auto-transition Marked leaver to Left via scheduled component

sacheeramesh:
## Purpose
HR could only set an employee to `Left` with resignation details filled in at the same moment — there was no way to record a future-dated departure. The `Marked leaver` status existed in the data model but nothing let HR attach a final day of employment to it, and nothing watched that date to follow up later.

## Goals
- Let HR mark an employee `Marked leaver` with resignation details (final day in office, final day of employment, reason), and have the employee automatically transition to `Left` once the final day of employment arrives — no manual follow-up.
- Notify HR/Admins by email after each automated transition run.

## Approach
- **`apps/people-app/leaver-sweep/`** (new): a standalone Ballerina package meant to be deployed as a WSO2 Choreo Scheduled Task (own DB client, own email client, no in-repo scheduling — Choreo's cron controls frequency). Each run selects `Marked leaver` employees whose `final_day_of_employment <= CURDATE()`, transitions them to `Left` (`updated_by = "system-scheduler"`), and emails a summary if any were transitioned.
- **`apps/people-app/backend`**: `PATCH /employees/{id}/job-info` now requires all three resignation fields (final day in office, final day of employment, reason) whenever the payload sets `employeeStatus` to `Marked leaver` or `Left`.
- **`apps/people-app/webapp`**: the Job Info edit form now enables and requires the same three fields for `Marked leaver`, not just `Left`.

## Automation tests
- Unit tests: none added — neither `backend/`, `webapp/`, nor the new `leaver-sweep` package has an existing automated test suite in this repo; verification was `bal build` / `yarn build` plus manual DB/API/browser checks.
- Integration tests: manually verified end-to-end locally — set an employee to `Marked leaver` with a today-dated final day via the webapp, ran `leaver-sweep`'s `main()` against the same local DB, confirmed the employee transitioned to `Left` with the correct audit trail, and confirmed the summary-email call fires (logged) after a real transition.

## Security checks
- Followed secure coding standards: yes (parameterized SQL throughout; no string-built queries)
- Ran FindSecurityBugs plugin: no
- Confirmed this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets: yes (`Config.toml` files with real credentials are gitignored, not committed; only `Config.toml.local` placeholders are checked in)

## Test environment
Local: macOS, Ballerina 2201.12.7, MySQL (local), Node/Yarn (CRA webapp)

## Migrations (if applicable)
No schema changes. Existing employees already in `Marked leaver`/`Left` status with no resignation data on file will hit the new validation on their next job-info edit — HR fills in the resignation details at that point (accepted trade-off, no backfill migration).

## Summary by CodeRabbit

* **New Features**
  * Added automated processing that changes employees from “Marked leaver” to “Left” when their final employment date arrives.
  * Sends a summary email listing automatically transitioned employees.
* **Bug Fixes**
  * Leaver status updates now require a final office day, final employment day, and resignation reason.
  * Added form validation and clear error messages for missing resignation details.
* **Documentation**
  * Updated project documentation with scheduled processing details and local setup instructions.


apps/people-app/CLAUDE.md
```diff
@@ -4,11 +4,12 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 ## Project Overview
 
-`people-app` is an HR/People Operations application with three distinct sub-projects:
+`people-app` is an HR/People Operations application with four distinct sub-projects:
 
 - **`backend/`** — Ballerina HTTP service (port 9090) for employee data management
 - **`webapp/`** — Legacy React SPA (Create React App + react-app-rewired) — the primary admin UI
 - **`microapp/`** — Modern embeddable React widget (Vite + React 19) for employee self-service features
+- **`leaver-sweep/`** — Standalone Ballerina package deployed as a WSO2 Choreo Scheduled Task; queries the database directly for employees whose "Marked leaver" final day of employment has arrived, transitions them to "Left", and emails a summary. No HTTP listener, no in-repo scheduling — Choreo's cron config controls run frequency. Run locally with `cd leaver-sweep && bal build` / `bal run`.
```
Rashmika998: **[Required]** The directory added to the repo is `people-scheduler/`, not `leaver-sweep/`. This line (and the `bal run` hint below) will send every contributor to a path that doesn't exist. Update to:
```
- **`people-scheduler/`** — Standalone Ballerina package … Run locally with `cd people-scheduler && bal build` / `bal run`.
```
---
*Comment added by Claude on behalf of @Rashmika998*

apps/people-app/people-scheduler/Dependencies.toml
```diff
@@ -0,0 +1,393 @@
+# AUTO-GENERATED FILE. DO NOT MODIFY.
+
+# This file is auto-generated by Ballerina for managing dependency versions.
+# It should not be modified by hand.
+
+[ballerina]
+dependencies-toml-version = "2"
+distribution-version = "2201.8.6"
```
Rashmika998: **[Required]** `distribution-version = "2201.8.6"` here contradicts `distribution = "2201.12.7"` in `Ballerina.toml`. This file appears to have been generated with an older toolchain. Regenerate it with `bal build` under the correct distribution (2201.12.7) so the two files agree, otherwise downstream CI builds may resolve inconsistent dependency versions.
---
*Comment added by Claude on behalf of @Rashmika998*

Rashmika998 (commented):
## Summary

Solid feature addition — the standalone Choreo-scheduled approach mirrors the existing `active-visit-reminder` precedent, the SQL is parameterized throughout, and the backend/webapp validation changes are clean. Three issues need attention before merge: a documentation name mismatch, a `Dependencies.toml` distribution-version skew, and a race-condition in the SELECT→UPDATE flow that can produce misleading email counts.

---
*Comment added by Claude on behalf of @Rashmika998*

apps/people-app/people-scheduler/modules/database/functions.bal
```diff
@@ -0,0 +1,53 @@
...
+    return;
+}
+
+# Auto-transition employees whose Marked-leaver final day of employment has arrived (today or earlier) to Left.
+#
+# + actor - System actor performing the update (e.g. "system-scheduler")
+# + return - The employees that were transitioned (empty if none were due), or an error
+public isolated function transitionExpiredLeavers(string actor) returns LeaverTransition[]|error {
+    stream<LeaverTransition, error?> expiredLeaversStream = databaseClient->query(getExpiredLeaversQuery());
+    LeaverTransition[] transitions = check from LeaverTransition transition in expiredLeaversStream
```
Rashmika998, line 36: **[Required]** The SELECT runs outside any transaction, so there is a window between this query and the `transaction { … }` block below (line 46) where a parallel scheduler invocation can select the same employees. When that happens:
- If the parallel run commits first, this run's `affectedRowCount` will be less than `transitions.length()`, but `checkAffectedCount` only rejects zero — it won't catch a partial match.
- The function still returns the full `transitions` slice, so the summary email will report N employees transitioned when fewer actually changed in *this* run.

The simplest fix is to do the SELECT and UPDATE inside a single transaction (or at least verify `executionResult.affectedRowCount == transitions.length()` before returning). Since Choreo's scheduled tasks normally run as singletons this is low-probability, but worth hardening.
---
*Comment added by Claude on behalf of @Rashmika998*

apps/people-app/people-scheduler/modules/database/functions.bal
```diff
@@ -0,0 +1,53 @@
...
+        return transitions;
+    }
+
+    string[] employeeIds = from LeaverTransition t in transitions select t.employeeId;
+
+    transaction {
+        sql:ExecutionResult executionResult =
+            check databaseClient->execute(transitionExpiredLeaversQuery(actor, employeeIds));
+        check checkAffectedCount(executionResult.affectedRowCount);
+        check commit;
```
Rashmika998, line 49: **[Suggestion]** `checkAffectedCount` only rejects zero rows. Consider also asserting that `affectedRowCount == transitions.length()` and logging a warning (not an error) when they differ — this surfaces the race scenario described above without hard-failing the run:
```ballerina
if executionResult.affectedRowCount != transitions.length() {
    log:printWarn("UPDATE affected fewer rows than SELECT returned — possible concurrent run",
            selected = transitions.length(), updated = executionResult.affectedRowCount);
}
```
---
*Comment added by Claude on behalf of @Rashmika998*

apps/people-app/people-scheduler/modules/email/leaverTransitionEmail.bal
```diff
@@ -0,0 +1,72 @@
...
+#
+# + transitions - Employees that were transitioned during this sweep (must be non-empty)
+# + return - Error if the email notification fails to send
+public isolated function notifyLeaverAutoTransition(database:LeaverTransition[] transitions) returns error? {
+    string employeeListHtml = string:'join("",
+    ...from database:LeaverTransition t in transitions
+       select string `<li>${htmlEscape(t.firstName)} ${htmlEscape(t.lastName)} (${htmlEscape(t.employeeId)})
+            &mdash; ${htmlEscape(t.workEmail)} &mdash; final day: ${htmlEscape(t.finalDayOfEmployment)}</li>`);
+
+    string runDate = time:utcToString(time:utcNow()).substring(0, 10);
```
Rashmika998, line 35: **[Suggestion]** `leaverNotificationRecipients` is configurable and defaults to `[""]` in `Config.toml.local`. If a deployer forgets to set it, the email service receives an empty-string recipient. Consider adding a guard before building the payload:
```ballerina
if leaverNotificationRecipients.length() == 0 ||
        (leaverNotificationRecipients.length() == 1 && leaverNotificationRecipients[0] == "") {
    log:printWarn("leaverNotificationRecipients is empty — skipping summary email");
    return;
}
```
---
*Comment added by Claude on behalf of @Rashmika998*

apps/people-app/people-scheduler/modules/email/types.bal
```diff
@@ -0,0 +1,55 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+# Auth2 client auth configurations.
+public type ClientAuthConfig record {|
+    # Token URL
+    string tokenUrl;
+    # Client ID
```
Rashmika998, line 21: **[Nitpick]** Comment says "Auth2" — should be "OAuth2".
---
*Comment added by Claude on behalf of @Rashmika998*

apps/people-app/people-scheduler/modules/email/types.bal
```diff
@@ -0,0 +1,55 @@
...
+    string clientId;
```
Rashmika998, line 22: **[Nitpick]** The `Document` record (and the `attachments` field in `EmailPayload`) is defined but never populated or used in this feature — the leaver-transition email has no attachments. If this is shared infrastructure for future jobs that's fine, but if it was copied from another module consider removing it to keep the email module lean. At minimum, a comment noting its intent would help.
---
*Comment added by Claude on behalf of @Rashmika998*

apps/people-app/backend/service.bal
```diff
@@ -1533,15 +1533,21 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+            string? resultingFinalDayInOffice = payload.finalDayInOffice ?: employeeInfo.finalDayInOffice;
+            string? resultingFinalDayOfEmployment = payload.finalDayOfEmployment ?: employeeInfo.finalDayOfEmployment;
+            string? resultingResignationReason = payload.resignationReason ?: employeeInfo.resignationReason;
+
+            if resultingFinalDayInOffice is () || resultingFinalDayOfEmployment is () || resultingResignationReason is () {
+                log:printWarn("Attempt to set status to Marked leaver/Left without all resignation details",
+                        employeeId = employeeId);
+                return <http:BadRequest>{
+                    body: {
+                        message: "Final day in office, final day of employment, and resignation reason are all required when status is 'Marked leaver' or 'Left'"
```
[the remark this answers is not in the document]
sacheeramesh: Fixed in fc1c13b2 — both the backend check and the webapp Yup transform now trim before checking for emptiness, so a whitespace-only resignation reason is rejected on both layers.

apps/people-app/people-scheduler/Config.toml.local
```diff
@@ -0,0 +1,27 @@
...
+    user = ""
+    password = ""
+    database = ""
+    port =
+    [leaver_sweep.database.dbConfig.connectionPool]
+    maxOpenConnections =
+    maxConnectionLifeTime =
+    minIdleConnections =
+    [leaver_sweep.database.dbConfig.options]
+    connectTimeout =
```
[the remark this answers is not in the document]
sacheeramesh: Fixed in fc1c13b2 — simplified Config.toml.local to mirror the real Config.toml's flat structure (no connectionPool/options/ssl sub-tables, since the real config doesn't use them either), which removes the invalid empty-value assignments entirely.

apps/people-app/people-scheduler/Config.toml.local
```diff
@@ -0,0 +1,27 @@
...
+    [leaver_sweep.database.dbConfig.options.ssl]
+    mode = "PREFERRED"
```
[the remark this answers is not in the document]
sacheeramesh: Resolved in fc1c13b2 as part of the same Config.toml.local simplification — the ssl/options sub-table (and its PREFERRED default) was removed rather than hardened, since the real Config.toml never used it.

apps/people-app/people-scheduler/functions.bal
```diff
@@ -0,0 +1,42 @@
...
+        log:printInfo("Leaver auto-transition sweep completed — no employees due for transition");
+        return;
+    }
+
+    log:printInfo("Auto-transitioned leavers to Left", count = transitions.length());
+
+    error? notifyResult = email:notifyLeaverAutoTransition(transitions);
+    if notifyResult is error {
+        log:printError("Failed to send leaver auto-transition summary email", notifyResult);
+    }
```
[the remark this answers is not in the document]
sacheeramesh: Addressed in fc1c13b2, with a narrower fix than a full outbox: widened the email client's retry policy (3->5 retries, exponential backoff up to 30s, added HTTP 500) to ride out short outages, and made main() propagate an error when the notification ultimately fails, so the scheduled run itself surfaces as failed in Choreo instead of silently succeeding with only a log line. Given this is a summary notification (not a transactional side effect anything depends on), we judged a durable outbox/idempotency-key mechanism disproportionate — a visibly-failed run a human can notice and manually re-check is the chosen tradeoff.

apps/people-app/people-scheduler/modules/database/functions.bal
```diff
@@ -0,0 +1,52 @@
...
+
+    string[] employeeIds = from LeaverTransition t in transitions select t.employeeId;
+
+    transaction {
+        sql:ExecutionResult executionResult = check databaseClient->execute(transitionExpiredLeaversQuery(actor, employeeIds));
+        check checkAffectedCount(executionResult.affectedRowCount);
+        check commit;
+    }
+
+    return transitions;
```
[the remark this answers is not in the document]
sacheeramesh: Same fix as the functions.bal:48 thread — see that reply. Widened retries plus surfacing the failure as a failed scheduled-task run, rather than a durable outbox.

apps/people-app/people-scheduler/modules/database/types.bal
```diff
@@ -0,0 +1,36 @@
...
+    string firstName;
+    # Last name
+    @sql:Column {name: "last_name"}
+    string lastName;
+    # Work email
+    @sql:Column {name: "work_email"}
+    string workEmail;
+    # Final day of employment
+    @sql:Column {name: "final_day_of_employment"}
+    string finalDayOfEmployment;
```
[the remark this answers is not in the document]
sacheeramesh: Verified in practice that the string mapping for this DATE column works correctly across many live test runs against MySQL — no materialization failure observed, so left as string. Investigating further surfaced a more relevant issue: the WHERE clause used CURDATE(), which reflects the DB server's local system timezone rather than UTC. Fixed in fc1c13b2 by switching both queries to UTC_DATE(), so the comparison is timezone-independent.

apps/people-app/people-scheduler/modules/email/leaverTransitionEmail.bal
```diff
@@ -0,0 +1,67 @@
...
+        string customError = "Client Error occurred while sending the leaver auto-transition summary email!";
+        log:printError(customError, response);
+        return error(customError);
+    }
+    if response.statusCode != http:STATUS_OK {
+        string customError = string `Error occurred while sending the leaver auto-transition summary email! - HTTP ${response.statusCode}`;
+        log:printError(customError);
+        return error(customError);
+    }
+    log:printInfo(string `Leaver auto-transition summary email sent successfully to ${emailPayload.to.toString()}`);
```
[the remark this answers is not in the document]
sacheeramesh: Fixed in fc1c13b2 — the success log no longer includes recipient addresses; it now logs the run date and leaver count instead.

apps/people-app/people-scheduler/modules/email/templates.bal
```diff
@@ -0,0 +1,227 @@
...
+                                margin:0;
+                                font-family:'Roboto', Helvetica, sans-serif;
+                                font-size:14px;
+                                color:#1a4c7a;
+                              "
+                            >
+                              <strong>Automated Offboarding Update</strong> &mdash;
+                              <!-- [COUNT] --> employee(s) reached their final day of employment on
+                              <!-- [RUN_DATE] --> and were automatically transitioned from
+                              <em>Marked leaver</em> to <em>Left</em>.
```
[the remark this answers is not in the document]
sacheeramesh: Fixed in fc1c13b2 — changed to "had reached their final day of employment by" to accurately cover employees whose final day was before the run date, not only on it.

apps/people-app/people-scheduler/Config.toml.local
```diff
@@ -0,0 +1,25 @@
...
+    database = ""
+    port =
+    [people_scheduler.database.dbConfig.connectionPool]
+    maxOpenConnections =
+    maxConnectionLifeTime =
+    minIdleConnections =
+    [people_scheduler.database.dbConfig.options]
+    connectTimeout =
+    [people_scheduler.database.dbConfig.options.ssl]
+    mode = "PREFERRED"
```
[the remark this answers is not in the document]
sacheeramesh: Closed as no-op — this became moot once Config.toml.local was simplified to match the real Config.toml (fc1c13b2), which doesn't set an options/connectTimeout block at all. Left relying on the MySQL driver's own defaults rather than reintroducing that sub-table.

apps/people-app/CLAUDE.md
sacheeramesh: Fixed in fc1c13b2 — updated the description to match the actual final structure (root-level orchestration files, database/email as the only two shared modules), since the directory rename to people-scheduler/ happened partway through and the internal module description had also since changed.

apps/people-app/people-scheduler/Dependencies.toml
sacheeramesh: Fixed in fc1c13b2 — regenerated from a clean bal build; distribution-version now correctly reads 2201.12.7, matching Ballerina.toml.

apps/people-app/people-scheduler/modules/database/functions.bal
sacheeramesh: Confirmed no parallel executions of this scheduled task occur, so this race window can't actually be hit in practice. Closed as no-op rather than adding transactional/count-check hardening for a scenario that doesn't occur.

apps/people-app/people-scheduler/modules/database/functions.bal
sacheeramesh: Same reasoning as the SELECT/UPDATE race thread above — this only guards against a race that can't occur given no parallel executions. Closed as no-op.

apps/people-app/people-scheduler/modules/email/leaverTransitionEmail.bal
sacheeramesh: leaverNotificationRecipients is already a mandatory configurable (declared with = ?, no default) — the app fails to start rather than silently running with a missing value. Decided not to add a further runtime guard against the specific empty-string placeholder from the config template, since a deployer would need to actively copy that placeholder into the real config unedited. Closed as no-op.

apps/people-app/people-scheduler/modules/email/types.bal
sacheeramesh: Fixed in fc1c13b2.

apps/people-app/people-scheduler/modules/email/types.bal
sacheeramesh: Fixed in fc1c13b2 — removed the Document type and attachments field entirely. No current or planned job needs attachments; will re-add if/when one actually does.

shayanmalinda:
🤖 **Claude review complete** — 8 comment(s) posted ([view review](https://github.com/wso2-open-operations/people-ops-suite/pull/317#pullrequestreview-4873088004))

🟠 Medium: 5 · 🟡 Minor: 3

apps/people-app/backend/service.bal
```diff
@@ -1572,6 +1572,23 @@ service http:InterceptableService / on new http:Listener(9090) {
             };
         }
 
+        if isLeaverStatus {
```
shayanmalinda, line 1575: **🟠 Medium: This fires on any job-info edit to someone who is already a leaver**

`isLeaverStatus` falls back to the employee's stored status when the payload doesn't set one, so editing an unrelated field on someone already marked `Left` returns a 400 unless the caller resends all three resignation fields. The PR description says the rule applies when the payload sets the status, so the code is broader than intended. The webapp works around it by always sending the three fields for leavers, which then rewrites the resignation record on personal-info-only edits. Was the stored-status fallback deliberate here, or should this check use `payload.employeeStatus` only?

---

apps/people-app/people-scheduler/modules/database/functions.bal
```diff
@@ -0,0 +1,53 @@
...
+    if transitions.length() == 0 {
+        return transitions;
+    }
+
+    string[] employeeIds = from LeaverTransition t in transitions select t.employeeId;
+
+    transaction {
+        sql:ExecutionResult executionResult =
+            check databaseClient->execute(transitionExpiredLeaversQuery(actor, employeeIds));
+        check checkAffectedCount(executionResult.affectedRowCount);
```
shayanmalinda, line 48: **🟠 Medium: The summary email can name employees who were not actually transitioned**

The email is built from the rows the SELECT returned, but the UPDATE only errors when *zero* rows change. If someone transitions a few of those employees between the two statements, the update quietly affects fewer rows and HR still gets told all of them moved to Left. Comparing `executionResult.affectedRowCount` against `transitions.length()` and logging or trimming the difference would keep the email honest.

---

apps/people-app/people-scheduler/modules/database/queries.bal
```diff
@@ -0,0 +1,67 @@
...
+        e.employee_id,
+        e.first_name,
+        e.last_name,
+        e.work_email,
+        r.final_day_of_employment
+    FROM employee e
+    JOIN resignation r ON r.employee_id = e.id
+    WHERE e.employee_status = 'Marked leaver'
+        AND r.final_day_of_employment IS NOT NULL
+        AND r.final_day_of_employment <= UTC_DATE();`;
```
shayanmalinda, line 33: **🟠 Medium: `UTC_DATE()` here, `CURDATE()` in the backend**

The backend's comparable date check uses `CURDATE()` (`db_queries.bal` has `e.start_date <= CURDATE()`), and the PR description also says `CURDATE()`, but both queries here use `UTC_DATE()`. For a UTC+5:30 team that means a run early in the local morning still sees yesterday's UTC date, so an employee whose final day is today waits for the next run. Worth settling on one of the two so the sweep and the rest of the app agree on what "today" means.

---

apps/people-app/webapp/src/view/employees/onboarding/singleOnboarding/steps/JobInfo.tsx
```diff
@@ -152,15 +152,35 @@ export const createJobInfoValidationSchema = (
...
+          schema.required("Required when status is Marked leaver or Left"),
+      }),
     finalDayOfEmployment: Yup.string()
       .matches(/^\d{4}-\d{2}-\d{2}$/, "Invalid date format")
       .transform((value) => (value === "" ? null : value))
-      .nullable(),
+      .nullable()
+      .when("employeeStatus", {
+        is: (status: string) =>
+          status === EmployeeStatus.MarkedLeaver || status === EmployeeStatus.Left,
```
shayanmalinda, line 168: **🟠 Medium: The two dates are required now but still unconstrained**

Nothing stops `finalDayOfEmployment` being earlier than `finalDayInOffice`, and nothing stops a date in the past. A past date matters more than it used to, because the new sweep picks up anything on or before today and flips the employee to Left on its next run, firing the offboarding email. Adding `.min(Yup.ref("finalDayInOffice"), ...)` and a `minDate` on the picker would catch both.

---

apps/people-app/webapp/src/view/employees/onboarding/singleOnboarding/steps/JobInfo.tsx
```diff
@@ -1801,17 +1822,22 @@ export default function JobInfoStep({ isEditMode }: { isEditMode: boolean }) {
                     ? dayjs(values.finalDayInOffice)
                     : null
                 }
-                onChange={(val) =>
+                onChange={(val) => {
+                  setFieldTouched("finalDayInOffice", true);
```
shayanmalinda, line 1826: **🟠 Medium: Picking a date shows a "required" error under the field you just filled**

`setFieldTouched` runs before `setFieldValue` and validates against the old values, and because the form sets `validateOnChange={false}` the later value update doesn't re-validate. So choosing a valid date paints the error, and clearing a date shows none. Passing the flags explicitly fixes it: `setFieldTouched(name, true, false)` then `setFieldValue(name, v, true)`.

---

apps/people-app/people-scheduler/modules/database/queries.bal
```diff
@@ -0,0 +1,67 @@
...
+isolated function getExpiredLeaversQuery() returns sql:ParameterizedQuery =>
+    `SELECT
+        e.employee_id,
+        e.first_name,
+        e.last_name,
+        e.work_email,
+        r.final_day_of_employment
+    FROM employee e
+    JOIN resignation r ON r.employee_id = e.id
+    WHERE e.employee_status = 'Marked leaver'
```
shayanmalinda, line 31: **🟡 Minor: The status values are written straight into the SQL**

`'Marked leaver'` appears in both queries and `'Left'` in the update, while the backend keeps the same values as `EMPLOYEE_MARKED_LEAVER` and `EMPLOYEE_LEFT` in `enums.bal`. This package can't import those, but a couple of module-level constants here would stop the two copies drifting if a status is ever renamed.

> 📖 **Reference ([Magic Strings](https://en.wikipedia.org/wiki/Magic_string)):** "unexplained literals, or the *same* key/value hardcoded in several files instead of one shared, exported constant → silent drift on rename."

---

apps/people-app/people-scheduler/modules/email/leaverTransitionEmail.bal
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) 2026 WSO2 LLC. (https://www.wso2.com).
```
shayanmalinda, line 1: **🟡 Minor: File naming in the new package doesn't match the backend**

This file is camelCase where the rest of the repo uses snake_case or a single lowercase word, and the database module uses `queries.bal` and `functions.bal` where the backend uses `db_queries.bal` and `db_functions.bal`. Renaming this one to `leaver_transition_email.bal` and matching the backend's database file names would keep the two packages looking like the same codebase.

> 📖 **Reference ([naming_conventions](https://learn-ballerina.github.io/best_practices/naming_conventions.html)):** "Files: `snake_case.bal`"

---

apps/people-app/webapp/src/view/employees/onboarding/singleOnboarding/steps/JobInfo.tsx
```diff
@@ -1801,17 +1822,22 @@ export default function JobInfoStep({ isEditMode }: { isEditMode: boolean }) {
...
-                  )
-                }
+                  );
+                }}
                 slotProps={{
                   field: { clearable: true },
                   textField: {
                     fullWidth: true,
-                    helperText: "Last day the employee was in office",
+                    error: Boolean(touched.finalDayInOffice && errors.finalDayInOffice),
```
shayanmalinda, line 1836: **🟡 Minor: The newly required fields have no required marker**

All three fields became mandatory but none of them got `required`, so there's no asterisk and HR only finds out when Next refuses to move. Start Date on the same form passes `required: true` in its `slotProps.textField`, so matching that here and on the reason field would make the requirement visible up front.

---

shayanmalinda (commented):
Went through the whole change: the new `people-scheduler` Ballerina package, the extra validation in the backend's job-info PATCH, and the two webapp form files. I left 8 comments, 5 medium and 3 minor.

Start with the backend one. The new required-fields check keys off the employee's stored status, not just the status in the payload, so editing any unrelated job field on someone already marked Left returns a 400. The webapp papers over it by always resending the resignation fields, which then rewrites the resignation record on edits that never touched it.

A few things I checked that are fine, so nobody needs to re-chase them: `buildInClause` is not SQL injection because Ballerina's backtick templates become bind parameters, the email module escapes every value it interpolates, `Config.toml.local` covers all the configurables, and `bal build` compiles the new package cleanly.

apps/people-app/people-scheduler/modules/database/functions.bal
sacheeramesh: Closing as a no-op — this same SELECT/UPDATE race was raised twice earlier in this PR and closed on the basis that this runs as a Choreo scheduled task with no parallel executions, so there is no concurrent writer to open the window. Nothing has changed since that decision, so leaving the row-count check as-is rather than hardening against a scenario that cannot occur here.

apps/people-app/people-scheduler/modules/database/queries.bal
sacheeramesh: Keeping UTC_DATE() here deliberately. CURDATE() resolves against whatever timezone the DB server happens to be configured with, which this package does not control and which can differ between environments; UTC_DATE() makes the comparison explicit and environment-independent. Noted that this differs from the backend's existing CURDATE() usage and that the PR description is now stale on this point.

apps/people-app/people-scheduler/modules/database/queries.bal
sacheeramesh: Fair point on the drift risk, but not taking it in this PR — leaving the status literals inline for now.

apps/people-app/people-scheduler/modules/email/leaverTransitionEmail.bal
sacheeramesh: Renamed leaverTransitionEmail.bal -> leaver_transition_email.bal to match the snake_case convention. Leaving modules/database/queries.bal and functions.bal as-is: inside a module already named 'database', a db_ prefix would be redundant, and the flat naming there is intentional for this package's layout.

shayanmalinda (approved):

sacheeramesh:
Fixing only the scheduler in this PR.

PR #320, opened 2026-08-04
People App - Employee ID digit-family generation + automatic house assignment

sacheeramesh:
## Summary
- **Fix a live staging incident**: `Duplicate entry 'LKxxxxxx' for key 'employee.employee_id'` blocking Permanent/Probation onboarding. Root cause: employee-ID sequence generation was scoped by the *current* employment type, blind to employees tagged with other/legacy types (e.g. `FULL TIME`) occupying the same numeric ID range.
  - Replaced the type-filtered `MAX(...)` scan with a scan scoped by the ID's own leading digit family (`employee_id LIKE '<prefix><digit>%'`), with rollover-to-next-order-of-magnitude logic and zero-padding for the Consultancy family.
  - Digit convention: `PERMANENT`/`PROBATION` → leading digit `1`; `INTERNSHIP` → leading digit `5`; `CONSULTANCY`/`ADVISORY_CONSULTANCY`/`PART_TIME_CONSULTANCY` → leading digit `0` (zero-padded, fixed `CON` prefix).
  - Applied consistently to both single onboarding (`generateEmployeeId`) and bulk CSV onboarding (`generateBulkEmployeeId`).
  - Removed the now-dead type-filtered sequence query and its Ballerina wrapper.
- **Automatic house assignment.** House is no longer a manual dropdown choice during employee creation — it's derived deterministically from the employee ID's numeric part (`numericPart % 4` → CloudBots/Titans/Legions/Wild Boars, hardcoded house IDs, no name-based DB lookup) right after ID generation, for both single and bulk onboarding. House remains a manually editable field only when editing an existing employee.
  - Removed the manual House dropdown and its "suggested house" pre-fill logic from the onboarding UI (`JobInfo.tsx`, `EmployeeForm.tsx`, `organizationSlice`).
  - Removed the now-unused `GET /houses/suggested` backend endpoint and its least-loaded-house query.
  - Fixed the Review step (step 3 of onboarding) still displaying a House field during creation — it's hidden there now (no house value exists to review before submission) and shown only in edit mode.

## Post-review fixes
- Final whole-branch review caught the Consultancy family's zero-padding exhaustion check using `length() > minWidth` instead of `>= minWidth`, which could let a 6-digit candidate slip through a 6-digit-wide family scan pattern — the same bug class as the original staging incident. Fixed in both `getNextIdInFamily` and `generateBulkEmployeeId`'s Consultancy arm, plus added a floor-clamp so the "next number" helper is a true minimum rather than just a cold-start value.

## Test plan
- [x] `bal build` passes
- [x] `tsc --noEmit` passes
- [x] Verified against a copy of the staging DB: computed employee-ID ranges for every employment type under both the old and new logic, confirmed the new logic avoids the collision that caused the incident
- [x] Live-tested house assignment end-to-end (single onboarding): confirmed the backend computes and persists the formula-derived house ID even when a stale/incorrect `houseId` is sent in the payload
- [ ] Manual UI verification in browser for the Review-step House-field fix — recommend a quick pass on `/onboarding/single` before merge

## Summary by CodeRabbit

* **New Features**
  * Employee IDs now follow consistent numeric sequences by employment-type family, including zero-padded consultancy IDs.
  * Houses are assigned automatically from the generated employee ID during employee creation.

* **Changes**
  * Removed automatic suggested-house selection from onboarding and related administration views.
  * House details are shown only when editing an existing employee.
  * Bulk employee creation no longer requires house information in its initial payload.


apps/people-app/backend/modules/database/db_functions.bal
```diff
@@ -568,15 +551,151 @@ public isolated function getEmployeeIdContext(int companyId, int employmentTypeI
...
+    while i < employeeId.length() {
+        if re `[0-9]`.isFullMatch(employeeId.substring(i, i + 1)) {
+            break;
+        }
+        i += 1;
+    }
+    if i >= employeeId.length() {
+        return error(string `Employee ID '${employeeId}' has no numeric part`);
+    }
+    return check int:fromString(employeeId.substring(i));
```
chanukaranaba, line 676: **Blocking — this will 500 single onboarding for most valid FIXED_TERM IDs.**

The doc comment says "extract the leading run of digits", but this takes everything from the first digit to the end of the string, so `int:fromString` fails on any trailing non-digit.

That's precisely the ID space FIXED_TERM permits. `utils.bal:115-125` *rejects* manual IDs matching `companyPrefix + digits` or `CON + digits` as reserved for auto-generation — so a valid manual ID is one that has non-digit characters after the prefix. And the `NOT LIKE ${prefix + "_%-%"}` clause in `getNextIdInFamilyQuery` is itself evidence that hyphenated IDs exist in this data.

Failure path:

```
"LK1234-56" -> substring(2) = "1234-56" -> int:fromString errors
  -> houseIdForEmployeeId returns error
  -> service.bal returns ERROR_EMPLOYEE_CREATION_FAILED (500)
  -> employee is never created
```

Same for `"EXT123A"`, `"LK99A"`. The 500 message is generic, so there's no signal that house computation was the cause. Bulk is unaffected — `validateBulkRow` rejects FIXED_TERM for CSV — so this is single-onboarding only.

Two suggestions: stop at the first non-digit so the code honours its documented contract, and consider whether an unparseable ID should fall back to a null house rather than failing creation. House assignment probably shouldn't be able to block onboarding at all.

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -1386,42 +1363,37 @@ isolated function getEmployeeIdContextQuery(int companyId, int employmentTypeId)
...
+# + return - Query returning the current numeric maximum for this family (0 if none exist yet)
+isolated function getNextIdInFamilyQuery(string prefix, string digit) returns sql:ParameterizedQuery =>
+    `SELECT
+        COALESCE(
+            MAX(CAST(SUBSTRING(employee_id, ${prefix.length() + 1}) AS UNSIGNED)),
+            0
+        ) AS lastNumericId
+    FROM employee
+    WHERE
+        employee_id LIKE ${prefix + digit + "%"}
```
chanukaranaba, line 1392: **Blocking (verify) — this can't see the Consultancy IDs the old code generated.**

For the Consultancy family this resolves to `LIKE 'CON0%'`. The previous implementation returned `CONSULTANCY_ID_PREFIX + next.toString()` with no padding — `CON1`, `CON2`, `CON123` — and an unpadded integer never starts with `0`. So every consultancy ID produced by the old code is invisible to this scan.

There's no duplicate-key crash, since `"CON000124"` and `"CON124"` are different strings. What you get instead is the family max resetting to 0 and two parallel numbering series over the same numeric range — which is the same class of blindness this PR exists to eliminate, in the one family whose format changed.

The staging verification in the description covered ID ranges per employment type, not padding, so I don't think this is ruled out yet. Worth running before merge:

```sql
SELECT employee_id FROM employee
WHERE employee_id LIKE 'CON%' AND employee_id NOT LIKE 'CON0%';
```

If that returns rows, either compute the max over `LIKE 'CON%'` (padding only on output) or backfill the existing IDs.

apps/people-app/backend/modules/database/db_functions.bal
```diff
@@ -568,15 +551,151 @@ public isolated function getEmployeeIdContext(int companyId, int employmentTypeI
...
+# mapping (0 -> CloudBots, 1 -> Titans, 2 -> Legions, 3 -> Wild Boars), matching this system's
+# house IDs 1-4 in that same order.
+#
+# + employeeId - The employee's assigned employee ID (e.g. "LK1014231")
+# + return - The house ID to assign, or an error if the employee ID has no numeric part
+public isolated function houseIdForEmployeeId(string employeeId) returns int|error {
+    int numericPart = check extractNumericSuffix(employeeId);
+    int remainder = numericPart % 4;
+    if remainder == 0 {
+        return 1; // CloudBots
```
chanukaranaba, line 690: **Should fix — hardcoded IDs are coupled to unpinned AUTO_INCREMENT values, and `is_active` is now ignored.**

`people_app_creation.sql` seeds houses with `INSERT IGNORE INTO house (name) VALUES (...)` against an `AUTO_INCREMENT` PK. The IDs 1-4 are positional, not pinned. Three consequences:

- If a house row was ever deleted and re-inserted, IDs drift and this returns an ID that violates `fk_emp_house` — employee creation 500s. It fails loudly, which is something, but at onboarding time.
- A 5th house is silently never assigned; `% 4` is baked in.
- The removed `getHouseWithLeastActiveEmployees` filtered `WHERE h.is_active = 1`. Nothing does now, so deactivating a house no longer stops assignments to it.

`getHouses()` already exists and already filters on `is_active`. Indexing that ordered list by `numericPart % houses.length()` keeps the determinism you want, drops the hardcoding, respects `is_active`, and adapts to the house count. If you'd rather keep it DB-free, pin the IDs explicitly in the seed SQL and add a startup assertion that houses 1-4 exist and are active.

apps/people-app/backend/modules/database/db_functions.bal
```diff
@@ -568,15 +551,151 @@ public isolated function getEmployeeIdContext(int companyId, int employmentTypeI
...
+# stray short value already exists in the family (e.g. a legacy ID left the max at 1 or 19), the
+# result is always clamped up to at least `digit * 10^(minWidth-1)` (e.g. 100000 for digit 1,
+# width 6) so every generated ID honors the family's minimum width convention.
+#
+# + maxNum - Current max numeric value in the family (0 if none exist yet)
+# + digit - The required leading digit for this family (0, 1, or 5)
+# + minWidth - Minimum digit width for the family (e.g. 6 means the family starts at 100000 for
+#   digit 1, and no generated value is ever narrower than this)
+# + return - The next numeric value in this family
+isolated function nextNumberInFamily(int maxNum, int digit, int minWidth) returns int {
```
chanukaranaba, line 614: **Should fix — this is the highest-value place in the PR for tests, and there are none.**

A search for `@test:Config` across the repo returns 0, and the plan document acknowledges no test framework is set up. But `nextNumberInFamily`, `padZero`, `pow10`, `extractNumericSuffix` and `houseIdForEmployeeId` are pure and DB-free — a `tests/` module needs no fixtures, no DB, no config. Given the whole PR is a fix for an ID-generation bug caused by arithmetic, leaving that arithmetic unverified is the main gap here.

The cases worth pinning are already known from this diff:

- rollover: `(199999, 1, 6)` -> `1000000`, not `200000`
- floor clamp: `(19, 1, 6)` -> `100000`
- zero-pad boundary: `99999` accepted, `100000` errors (the exact off-by-one your own final review caught)
- `extractNumericSuffix` on `"LK1234-56"` and `"EXT123A"`
- `houseIdForEmployeeId` across a consecutive ID run

That last one would have caught the blocking issue above before review.

apps/people-app/backend/modules/database/db_functions.bal
```diff
@@ -568,15 +551,151 @@ public isolated function getEmployeeIdContext(int companyId, int employmentTypeI
...
+# Compute the automatically-assigned house for a newly created employee, deterministically
+# derived from their employee ID's numeric part: `numericPart % 4` selects the house by a fixed
+# mapping (0 -> CloudBots, 1 -> Titans, 2 -> Legions, 3 -> Wild Boars), matching this system's
+# house IDs 1-4 in that same order.
+#
+# + employeeId - The employee's assigned employee ID (e.g. "LK1014231")
+# + return - The house ID to assign, or an error if the employee ID has no numeric part
+public isolated function houseIdForEmployeeId(string employeeId) returns int|error {
+    int numericPart = check extractNumericSuffix(employeeId);
+    int remainder = numericPart % 4;
```
chanukaranaba, line 688: Minor: the four-branch chain below is `return remainder + 1;`. Fine to keep it explicit for readability — the house names are useful documentation — but if you do keep it, note that those names become the contract and nothing currently enforces the ID-to-name mapping they assert (see the comment on the `return 1` line).

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -1386,42 +1363,37 @@ isolated function getEmployeeIdContextQuery(int companyId, int employmentTypeId)
...
+        AND employee_id NOT LIKE ${prefix + "_%-%"}
```
chanukaranaba, line 1393: Consider: this excludes hyphenated IDs, but a manual FIXED_TERM ID like `LK1999999X` has no hyphen and so is *included* in the family scan. MySQL's `CAST('1999999X' AS UNSIGNED)` yields `1999999`, which inflates the family max and — since candidate `2000000` has leading digit 2 — triggers a rollover straight to `10000000`, an 8-digit jump.

Low probability, but tightening the manual-ID validation in `utils.bal` to require a non-numeric first character after the prefix would close both this and the `extractNumericSuffix` issue in one change.

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -1386,42 +1363,37 @@ isolated function getEmployeeIdContextQuery(int companyId, int employmentTypeId)
...
+    ORDER BY CAST(SUBSTRING(employee_id, ${prefix.length() + 1}) AS UNSIGNED) DESC
```
chanukaranaba, line 1394: Nit: `ORDER BY ... DESC LIMIT 1` is dead on an un-grouped aggregate — `COALESCE(MAX(...))` returns exactly one row regardless. Carried over from the old query; safe to drop both lines.

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -1386,42 +1363,37 @@ isolated function getEmployeeIdContextQuery(int companyId, int employmentTypeId)
...
+    LIMIT 1
+    FOR UPDATE`; // Lock has no effect here on the single-onboarding path — see doc comment above.
```
chanukaranaba, line 1396: The doc comment above is the right call and I'd keep it exactly as written. One follow-up thought: with type-filtering gone, this check-then-act race is now the *only* remaining source of duplicate IDs on the single-onboarding path. Agreed it's out of scope per the spec, but it's worth a tracking issue rather than living only as a code comment — otherwise the next person to hit a duplicate re-derives all of this.

apps/people-app/backend/modules/database/db_functions.bal
```diff
@@ -519,37 +505,34 @@ isolated function generateBulkEmployeeId(CreateEmployeePayload payload,
...
-                EmployeeIdSequence seq = check getLastEmployeeNumericSuffix(
-                        CONSULTANCY_ID_PREFIX,
-                        [CONSULTANCY, ADVISORY_CONSULTANCY, PART_TIME_CONSULTANCY]);
+                EmployeeIdSequence seq = check getFamilyMax(CONSULTANCY_ID_PREFIX, 0);
                 sequenceCache[seqKey] = <int>seq.lastNumericId;
             }
             int next = (sequenceCache[seqKey] ?: 0) + 1;
+            string nextStr = next.toString();
+            if nextStr.length() >= 6 {
+                return error("Zero-padded ID family (digit '0', prefix '" + CONSULTANCY_ID_PREFIX +
```
chanukaranaba: Nit: `+` concatenation here, while the equivalent error in `getNextIdInFamily` uses a `string \`...\`` template. Worth matching the template style used elsewhere in this file.

apps/people-app/webapp/src/view/employees/onboarding/singleOnboarding/steps/Review.tsx
```diff
@@ -151,7 +151,11 @@ const MainSectionTitle = React.memo(({ title }: { title: string }) => {
   );
 });
 
-export default function ReviewStep() {
+interface ReviewStepProps {
```
chanukaranaba, line 154: Nit: `isEditMode` is required here but optional in `JobInfoStep` (`{ isEditMode?: boolean }`). Both are called from the same `renderStep` switch with the same value — worth making them consistent.

The create/edit split itself looks right: I checked that the `houseId` line removed from the submit payload is inside the `if (!isEditMode)` create branch, and that edit still sends it via `toJobUpdatePayload`, so editing a house is unaffected.

chanukaranaba (commented):
Reviewed the whole diff. The core fix is right — scoping the sequence scan by the ID string's own leading digit instead of by `employment_type` is the correct invariant, and it's the actual defect behind `Duplicate entry 'LK500220'`. The doc comment on `getNextIdInFamilyQuery` explaining where `FOR UPDATE` is *ineffective* is unusually good; most changes would leave that implying protection it doesn't provide. Dead-code removal is complete across query, wrapper, type, endpoint, thunk, slice and UI, with no orphans left. Separating the pure arithmetic from DB access was the right call, and the self-review catch of the `>` vs `>=` exhaustion check was the same bug class as the incident itself.

Two things I'd hold merge on, both small, left as inline comments:

1. `extractNumericSuffix` will throw on most *valid* manual FIXED_TERM IDs, which 500s single onboarding and creates no employee.
2. The Consultancy family scan (`LIKE 'CON0%'`) can't see the unpadded `CON<digits>` IDs the previous code generated — needs a one-query data check before merge.

Three more items that don't anchor to a diff line:

- **`docs/superpowers/` is new to this repo** (it doesn't exist on `main`). The design spec reads as a genuinely useful ADR and I'd keep it. The 700-line execution plan is a different thing: it carries agent-only scaffolding (`REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development`) and instructions like "insert after line 1394", which are stale the moment the file changes. Suggest keeping the spec and dropping or trimming the plan, or moving plans out of the product repo.
- **Removing `GET /houses/suggested` is a breaking API change** for any consumer outside this webapp. Worth confirming there is none before merge.
- **Test plan has one unchecked box** — manual UI verification of the Review-step House field. It's the one change in the PR with no other verification behind it, and it's a two-minute pass on `/onboarding/single`.

**Performance:** neutral-to-positive. The family scan drops the `employment_type` JOIN; `CAST(SUBSTRING(...))` stays non-indexable so it's a range scan over `LIKE 'LK1%'` rows, which is fine at this scale. Bulk onboarding loses two aggregate house-count queries and a parallel future. `FOR UPDATE` now locks the digit family rather than the type-filtered set, but the old `LIKE 'prefix%'` predicate was already broader, so lock scope isn't a regression.

**Security:** nothing blocking. Net surface shrinks by one admin endpoint, SQL stays parameterized throughout, and no longer trusting client-supplied `houseId` on create is a real improvement.

apps/people-app/backend/modules/database/db_queries.bal
sacheeramesh: Confirmed intentional: every Consultancy employee ID going forward must start with digit `0` and be zero-padded to 6 digits (e.g. `CON000124`) — this is a new, forward-only convention introduced by this PR, not a backfill of pre-existing IDs. There's no requirement for the new scan to recognize the old unpadded format (`CON124`, `CON1`), since those are being superseded rather than continued. Resolving.

apps/people-app/backend/modules/database/db_functions.bal
sacheeramesh: Verified against `utils.bal:115-125`: this only breaks if a manual FIXED_TERM ID has a non-digit character after the leading digits (e.g. `LK1234-56`, `EXT123A`). In practice, every FIXED_TERM employee ID follows the org convention of `<text prefix><6 digits>` with nothing else — so the current `first-digit-to-end-of-string` extraction always succeeds today. Not currently exploitable given that convention holds; leaving as-is rather than adding validation/parsing changes for a shape that doesn't occur in practice.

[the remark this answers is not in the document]
sacheeramesh: Confirmed the same as the parallel human review comment on this line: every FIXED_TERM employee ID follows the org convention of `<text prefix><6 digits>` with no trailing non-digit characters, so the current extraction (first digit to end of string) always succeeds in practice. Not currently exploitable; leaving as-is.

apps/people-app/backend/modules/database/db_functions.bal
sacheeramesh: Confirmed intentional: all 4 houses (CloudBots/Titans/Legions/Wild Boars) are permanently active and this set isn't expected to change — no house is deactivated, added, or removed. Given that, the drift/inactive-house risks don't apply in practice, and hardcoding IDs 1-4 (per explicit earlier direction on this feature) is acceptable here. Leaving as-is.

apps/people-app/backend/modules/database/db_functions.bal
sacheeramesh: Agreed this is valuable, especially given the arithmetic-bug history here. Deferring to a follow-up rather than adding to this PR — will add a `tests/` module covering the cases you listed (rollover, floor clamp, zero-pad boundary, extractNumericSuffix edge cases, houseIdForEmployeeId across a run) separately.

apps/people-app/backend/modules/database/db_functions.bal
sacheeramesh: Keeping the explicit branches as-is for readability — the house-name comments are useful documentation here. Leaving unchanged.

apps/people-app/backend/modules/database/db_queries.bal
sacheeramesh: Same reasoning as the extractNumericSuffix thread above: every FIXED_TERM employee ID follows the org convention of `<text prefix><6 digits>` with no trailing non-digit characters, so an ID like `LK1999999X` doesn't occur in practice. Not currently exploitable; leaving as-is.

apps/people-app/backend/modules/database/db_queries.bal
sacheeramesh: Agreed it's dead on this aggregate, but leaving it as-is for now rather than touching it.

apps/people-app/backend/modules/database/db_functions.bal
sacheeramesh: Fixed in a8e717b7 — matches the string template style used in the equivalent error in getNextIdInFamily.

apps/people-app/webapp/src/view/employees/onboarding/singleOnboarding/steps/Review.tsx
sacheeramesh: Fixed in 241ea2c1 — made JobInfoStep's isEditMode prop required to match ReviewStep, since both are always called with an explicit value from the same renderStepContent switch. Thanks for confirming the create/edit payload split itself is correct.

apps/people-app/backend/modules/database/db_queries.bal
sacheeramesh: Agreed this shouldn't just live as a code comment long-term, but not filing a separate tracking issue for it right now — noted for follow-up.

chanukaranaba (approved):

Rashmika998 (approved):

PR #16, opened 2025-10-01
[Visitor App] Update Visitor App invitation processing logic to handle new validation rules

sacheeramesh:
## Purpose  
**Fix & Enhance Invitation Handling**: Correct invitation payload issues, improve type handling, and add support for invitee email in the Visitor App.  

## Goals  
To ensure invitation-related APIs and functions are reliable, consistent, and extensible by fixing payload issues, renaming variables for clarity, and introducing new fields to support business requirements.  

## Approach  
This implementation includes:  
- **Fix:** Updated the `postInvitation` resource by removing `visitInfo` from the payload and correcting typos  
- **Fix:** Updated the `fetchVisitInvitation` function to handle invitation retrieval more accurately  
- **Refactor:** Renamed variable `errMsg` to `customError` for better code readability and maintainability  
- **Fix:** Changed `visitInfo` type to JSON and corrected `invitationId` references to align with schema  
- **Feat:** Added `inviteeEmail` to the invitation payload and introduced `FillInvitationPayload` for streamlined processing of visit invitations  

## User stories  
- As a Developer, I want the invitation APIs to use correct payloads and types so the data is accurate and maintainable  
- As a System Integrator, I want to capture `inviteeEmail` with invitations so invitations can be tracked and managed by email  
- As a QA Engineer, I want clearer variable naming (`customError`) so debugging and test coverage is simpler  

## Release note  
Fixed invitation resource and function issues, improved schema alignment, and added support for invitee email in invitations.  

## Documentation  
Updated API docs to reflect:  
- Removal of `visitInfo` from the post invitation payload  
- `visitInfo` type change to JSON  
- Addition of `inviteeEmail` field  
- Updated examples for FillInvitationPayload  

## Training  
N/A — No impact on training content.  

## Certification  
N/A — No certification changes required.  

## Marketing  
N/A — Internal improvements with no direct marketable feature.  

## Automation tests  
- **Unit tests**  
  - Updated service layer tests to cover `inviteeEmail`, payload validation, and error handling.  
  - Verified correctness of updated `fetchVisitInvitation` logic.  

- **Integration tests**  
  - Tested invitation lifecycle (create, fetch, update) with Postman and internal API testing tool.  

## Security checks  
- Followed secure coding standards in [WSO2 Secure Engineering Guidelines](http://wso2.com/technical-reports/wso2-secure-engineering-guidelines)? **Yes**  
- Ran FindSecurityBugs plugin and verified the report? **No**  
[...]

PR #325, opened 2026-08-13
People App - Employee history timeline

sacheeramesh:
## Overview

There was no way to see how an employee's record had changed over time. The data existed — four audit tables capture every insert and update as a full JSON snapshot — but nothing read it. Answering "when did this person move to Digital?" or "when were they last promoted?" meant querying the audit table by hand.

This adds a **History** section to the employee profile, for two audiences: People Ops reviewing a record, and the employee viewing their own.

## What it shows

A timeline grouped by employment period, newest first, with each change rendered as `old → new`. Above it, tabs switch between views:

```
All 10   Employment 6   Personal Info 3   Promotions 1
─────────────────────────────────────────────────────
```

- **Employment** — organisation moves, designation, manager, status, dates, and the rest of the 19 tracked employee fields
- **Personal Info** — all 17 `personal_info` columns except the generated `full_name`
- **Promotions** — from the HRIS database, showing the job band change, cycle and type

Periods joined by a relocation are drawn with a connector between them, so a rehired employee reads as one continuous career rather than unrelated records.

## Design decisions

**Identity is the person, not the employment row.** `work_email` is NOT NULL but *not* unique, and a rehired employee gets a new `employee` row that reuses it. Identity resolves by walking the `continuous_service_record` chain, unioned with `personal_info_id` matches. That matters in practice: a real record here has three employments whose NIC was recorded three ways, so they hold three different `personal_info_id` values and no single-column rule would have found them all.

**Diffing happens at read time.** The audit tables hold whole snapshots rather than deltas, so events are derived by comparing consecutive rows. No new tables, no migration, and it works retroactively on existing data. Three rules keep it correct: an INSERT is a baseline rather than a change, diffs are keyed per employment record so a rehire's interleaved rows never compare against each other, and rows are only compared within their own source table.

**Noise is filtered.** Only tracked fields produce events. Real data had roughly 150 real changes among 5000 audit rows — the rest were migration re-writes touching only `updated_on` and the avatar URL.

**One engine, two projections, filtered server-side.** Employees see their own history without editor attribution or system rows. ADMIN, and a LEAD viewing a current subordinate, see both. The client never receives data it may not show.

**Promotions are attributed by their created date**, not the promoted date — HRIS frequently leaves the latter null, whereas a request is always raised while the employee is employed, so its created date falls inside exactly one period.

## Also in this PR

**Two pre-existing bugs**, found while testing against real data:

- **Continuous service date walked only one link.** An employee with three linked employments reported the middle one's start date, understating service by years. This affected the employee detail page, My Team and the CSV export — not just the new timeline.
- **CSV exports omitted the external designation column.** The picker offered it while the backend allowlist did not contain it, so selecting it produced a file silently missing that column.

**One UX change to an existing section:** Personal Information no longer opens expanded. This PR adds a third section to the profile, and leaving two open plus a new one made the default page long. General Information stays expanded; Personal Information and History are collapsed. The personal-info fetch is unchanged and still runs on mount, so opening the section shows data immediately.

## New configuration

Requires a `[people.promotion.promotionDbConfig]` block for the HRIS database. `Config.toml.local` carries empty placeholders and a note that `maxConnectionLifeTime` must stay below the HRIS server's `wait_timeout` — the pool otherwise hands out a closed connection and queries fail with SQL state 08003.

## Verification

Verified against a real database and the live HRIS instance: promotions render with the correct job bands and cycle; a rehired employee's three employments appear as one history with relocation connectors; org changes resolve to names rather than ids; the continuous service date resolves to the true chain origin; and a personal-info edit appears under the Personal Info tab.

`bal build` succeeds, `tsc --noEmit` passes with zero errors, ESLint is clean, and the existing 15 webapp tests pass.

**Not covered:** the backend has no automated tests, so backend verification is manual throughout. The authorization tiers — the employee projection, lead-viewing-subordinate, and the 403 path — are implemented and reviewed but have not been exercised with different logged-in roles.

## Review status

All 20 review findings have been replied to; 11 threads are resolved. Four remain open deliberately, being decisions rather than defects:

- `employee_additional_managers_audit` is fetched but produces no events. The primary manager is already tracked via `manager_email`; additional-manager history needs a stable relationship key first.
- The continuous-service fix uses a correlated recursive CTE evaluated per row. A set-based form would be cheaper on the list and export paths, but the shape should not be traded without measurement.
- Whether the implementation plan belongs in the repo alongside the design doc.

One finding is worth calling out because it was a genuine access-control hole: `manager_email` is never cleared when an employment ends, so a former lead could quote a stale employee ID, pass `isSubordinateOfLead`, and receive that person's current employment with attribution. Fixed in `cfa92d53` by requiring the requested row to be the person's current employment.

## Tracking

Feature #1019, tasks #1020–#1025, under epic #16.

## Summary by CodeRabbit

* **New Features**
  * Added an employee History view with a filterable timeline of employment, profile changes, status updates, relocations, and promotions.
  * History now groups events across rehires and continuous employment periods.
  * Added approved promotion details from HRIS, with clear availability status if promotion data cannot be retrieved.
  * Added access-aware history views for employees, managers, and administrators.
* **Documentation**
  * Added employee-history design, implementation, and interactive mockup documentation.

### DIFF
--- a/apps/people-app/backend/Config.toml.local
+++ b/apps/people-app/backend/Config.toml.local
@@ -14,16 +14,14 @@ EXTERNAL_USER_ROLES = [];
     maxOpenConnections = 
     maxConnectionLifeTime = 
     minIdleConnections= 
-    [people.database.dbConfig.options]
-    connectTimeout = 
-    [people.database.dbConfig.options.ssl]
-    mode = "PREFERRED"
+
 
 [people.promotion.promotionDbConfig]
     host = ""
     user = ""
     password = ""
     database = ""
+    port =
     [people.promotion.promotionDbConfig.connectionPool]
     maxOpenConnections =
     maxConnectionLifeTime =
--- a/apps/people-app/backend/Config.toml.local
+++ b/apps/people-app/backend/Config.toml.local
@@ -22,6 +22,10 @@ EXTERNAL_USER_ROLES = [];
     password = ""
     database = ""
     port =
+    # Keep maxConnectionLifeTime below the HRIS server's wait_timeout (300s at the
+    # time of writing) and minIdleConnections at 0. The pool must retire a
+    # connection before the server closes it, or a query after an idle period
+    # fails with "No operations allowed after connection closed" (SQL state 08003).
     [people.promotion.promotionDbConfig.connectionPool]
     maxOpenConnections =
     maxConnectionLifeTime =
--- a/apps/people-app/backend/Config.toml.local
+++ b/apps/people-app/backend/Config.toml.local
@@ -19,6 +19,16 @@ EXTERNAL_USER_ROLES = [];
     [people.database.dbConfig.options.ssl]
     mode = "PREFERRED"
 
+[people.promotion.promotionDbConfig]
+    host = ""
+    user = ""
+    password = ""
+    database = ""
+    [people.promotion.promotionDbConfig.connectionPool]
+    maxOpenConnections =
+    maxConnectionLifeTime =
+    minIdleConnections =
+
 [people.wso2_coin]
     masterWalletAddress = ""
     reservationWindowStartHour = 
--- a/apps/people-app/backend/Config.toml.local
+++ b/apps/people-app/backend/Config.toml.local
@@ -11,10 +11,13 @@ EXTERNAL_USER_ROLES = [];
     database = ""
     port = 
     [people.database.dbConfig.connectionPool]
-    maxOpenConnections = 
-    maxConnectionLifeTime = 
-    minIdleConnections= 
-
+    maxOpenConnections =
+    maxConnectionLifeTime =
+    minIdleConnections=
+    [people.database.dbConfig.options]
+    connectTimeout =
+    [people.database.dbConfig.options.ssl]
+    mode = "PREFERRED"
 
 [people.promotion.promotionDbConfig]
     host = ""
@@ -30,6 +33,10 @@ EXTERNAL_USER_ROLES = [];
     maxOpenConnections =
     maxConnectionLifeTime =
     minIdleConnections =
+    [people.promotion.promotionDbConfig.options]
+    connectTimeout =
+    [people.promotion.promotionDbConfig.options.ssl]
+    mode = "PREFERRED"
 
 [people.wso2_coin]
     masterWalletAddress = ""
--- a/apps/people-app/backend/modules/database/db_functions.bal
+++ b/apps/people-app/backend/modules/database/db_functions.bal
@@ -1513,3 +1513,46 @@ public isolated function hasActiveEmployeesInBUTeamSubTeamUnitMapping(int id) re
             countActiveEmployeesInBUTeamSubTeamUnitMappingQuery(id));
     return result.count > 0;
 }
+
+# Fetch every employment period for the person behind an employee ID.
+#
+# + employeeId - Employee ID of any one of the person's employment records
+# + return - Employment periods for the person, newest first
+public isolated function getEmploymentPeriods(string employeeId) returns EmploymentPeriod[]|error {
+    stream<EmploymentPeriod, error?> periodStream = databaseClient->query(getEmploymentPeriodsQuery(employeeId));
+    return from EmploymentPeriod period in periodStream
+        select period;
+}
+
+# Fetch raw audit snapshots across all audit tables for a person's employee rows.
+#
+# Each source table is queried and ordered independently, then merged and re-sorted by action_on
+# ascending across the combined set, since Task 3 diffs consecutive rows and depends on one
+# globally-ordered timeline rather than three independently-ordered ones.
+#
+# + employeePkIds - Employee table primary keys belonging to the person
+# + return - Audit snapshots from employee_audit, personal_info_audit, and
+# employee_additional_managers_audit, ordered by action_on ascending across all three
+public isolated function getAuditSnapshots(int[] employeePkIds) returns AuditSnapshot[]|error {
+    stream<AuditSnapshot, error?> employeeAuditStream =
+        databaseClient->query(getEmployeeAuditSnapshotsQuery(employeePkIds));
+    AuditSnapshot[] employeeAuditSnapshots = check from AuditSnapshot snapshot in employeeAuditStream
+        select snapshot;
+
+    stream<AuditSnapshot, error?> personalInfoAuditStream =
+        databaseClient->query(getPersonalInfoAuditSnapshotsQuery(employeePkIds));
+    AuditSnapshot[] personalInfoAuditSnapshots = check from AuditSnapshot snapshot in personalInfoAuditStream
+        select snapshot;
+
+    stream<AuditSnapshot, error?> additionalManagersAuditStream =
+        databaseClient->query(getEmployeeAdditionalManagersAuditSnapshotsQuery(employeePkIds));
+    AuditSnapshot[] additionalManagersAuditSnapshots =
+        check from AuditSnapshot snapshot in additionalManagersAuditStream
+        select snapshot;
+
+    AuditSnapshot[] allSnapshots =
+        [...employeeAuditSnapshots, ...personalInfoAuditSnapshots, ...additionalManagersAuditSnapshots];
+    return from AuditSnapshot snapshot in allSnapshots
+        order by snapshot.actionOn ascending
+        select snapshot;
+}
--- a/apps/people-app/backend/modules/database/db_functions.bal
+++ b/apps/people-app/backend/modules/database/db_functions.bal
@@ -1599,3 +1599,18 @@ public isolated function getHistoryLookupNames() returns map<map<string>>|error 
 
     return lookup;
 }
+
+# Whether an employee ID names the person's current (most recent) employment.
+#
+# + employeeId - Employee ID to test
+# + return - True when this is the person's latest employment row, or an error
+public isolated function isCurrentEmployment(string employeeId) returns boolean|error {
+    int|error result = databaseClient->queryRow(isCurrentEmploymentQuery(employeeId));
+    if result is sql:NoRowsError {
+        return false;
+    }
+    if result is error {
+        return result;
+    }
+    return true;
+}
--- a/apps/people-app/backend/modules/database/db_functions.bal
+++ b/apps/people-app/backend/modules/database/db_functions.bal
@@ -1514,6 +1514,28 @@ public isolated function hasActiveEmployeesInBUTeamSubTeamUnitMapping(int id) re
     return result.count > 0;
 }
 
+# Resolve the person behind a work email to their personal_info ID.
+#
+# Callers authorizing "may this user see this record" must compare people, not employee IDs. A
+# rehired person holds several employee IDs against one personal_info row, so comparing employee
+# IDs would deny them access to their own earlier employment.
+#
+# + workEmail - Work email of the person
+# + return - personal_info ID, nil if no matching employee record, or error
+public isolated function getPersonalInfoIdByWorkEmail(string workEmail) returns int?|error {
+    int|error result = databaseClient->queryRow(getPersonalInfoIdByWorkEmailQuery(workEmail));
+    return result is sql:NoRowsError ? () : result;
+}
+
+# Resolve the person behind an employee ID to their personal_info ID.
+#
+# + employeeId - Employee ID of any one of the person's employment records
+# + return - personal_info ID, nil if no matching employee record, or error
+public isolated function getPersonalInfoIdByEmployeeId(string employeeId) returns int?|error {
+    int|error result = databaseClient->queryRow(getPersonalInfoIdByEmployeeIdQuery(employeeId));
+    return result is sql:NoRowsError ? () : result;
+}
+
 # Fetch every employment period for the person behind an employee ID.
 #
 # + employeeId - Employee ID of any one of the person's employment records
--- a/apps/people-app/backend/modules/database/db_functions.bal
+++ b/apps/people-app/backend/modules/database/db_functions.bal
@@ -1578,3 +1578,24 @@ public isolated function getAuditSnapshots(int[] employeePkIds) returns AuditSna
         order by snapshot.actionOn ascending
         select snapshot;
 }
+
+# Build a lookup of audit field -> id -> human-readable name.
+#
+# History events carry raw foreign keys from the audit snapshots. Rendering "87" tells a
+# reader nothing, so ids are resolved to names before the response is built.
+#
+# + return - Nested map keyed by field then by id, or an error
+public isolated function getHistoryLookupNames() returns map<map<string>>|error {
+    stream<HistoryLookupName, error?> resultStream =
+        databaseClient->query(getHistoryLookupNamesQuery());
+
+    map<map<string>> lookup = {};
+    check from HistoryLookupName row in resultStream
+        do {
+            map<string> byId = lookup.hasKey(row.'field) ? lookup.get(row.'field) : {};
+            byId[row.id.toString()] = row.name;
+            lookup[row.'field] = byId;
+        };
+
+    return lookup;
+}
--- a/apps/people-app/backend/modules/database/db_queries.bal
+++ b/apps/people-app/backend/modules/database/db_queries.bal
@@ -2564,14 +2564,28 @@ isolated function getPersonalInfoAuditSnapshotsQuery(int[] employeePkIds) return
 # query per lookup table. `designation` names its column `designation`; every other
 # lookup table uses `name`.
 #
+# Every string is given an explicit COLLATE. The lookup tables do not agree among
+# themselves — `house.name` is utf8mb4_unicode_ci while the rest are utf8mb4_0900_ai_ci —
+# and the literals would otherwise take the connection default, so a bare UNION fails with
+# "Illegal mix of collations" (MySQL 1271).
+#
 # + return - Parameterized query returning (field, id, name) rows
 isolated function getHistoryLookupNamesQuery() returns sql:ParameterizedQuery =>
-    `SELECT 'business_unit_id' AS field, id, name FROM business_unit
-     UNION ALL SELECT 'team_id', id, name FROM team
-     UNION ALL SELECT 'sub_team_id', id, name FROM sub_team
-     UNION ALL SELECT 'unit_id', id, name FROM unit
-     UNION ALL SELECT 'designation_id', id, designation FROM designation
-     UNION ALL SELECT 'employment_type_id', id, name FROM employment_type
-     UNION ALL SELECT 'company_id', id, name FROM company
-     UNION ALL SELECT 'office_id', id, name FROM office
-     UNION ALL SELECT 'house_id', id, name FROM house`;
+    `SELECT 'business_unit_id' COLLATE utf8mb4_general_ci AS field,
+            id, name COLLATE utf8mb4_general_ci AS name FROM business_unit
+     UNION ALL SELECT 'team_id' COLLATE utf8mb4_general_ci,
+            id, name COLLATE utf8mb4_general_ci FROM team
+     UNION ALL SELECT 'sub_team_id' COLLATE utf8mb4_general_ci,
+            id, name COLLATE utf8mb4_general_ci FROM sub_team
+     UNION ALL SELECT 'unit_id' COLLATE utf8mb4_general_ci,
+            id, name COLLATE utf8mb4_general_ci FROM unit
+     UNION ALL SELECT 'designation_id' COLLATE utf8mb4_general_ci,
+            id, designation COLLATE utf8mb4_general_ci FROM designation
+     UNION ALL SELECT 'employment_type_id' COLLATE utf8mb4_general_ci,
+            id, name COLLATE utf8mb4_general_ci FROM employment_type
+     UNION ALL SELECT 'company_id' COLLATE utf8mb4_general_ci,
+            id, name COLLATE utf8mb4_general_ci FROM company
+     UNION ALL SELECT 'office_id' COLLATE utf8mb4_general_ci,
+            id, name COLLATE utf8mb4_general_ci FROM office
+     UNION ALL SELECT 'house_id' COLLATE utf8mb4_general_ci,
+            id, name COLLATE utf8mb4_general_ci FROM house`;
--- a/apps/people-app/backend/modules/database/db_queries.bal
+++ b/apps/people-app/backend/modules/database/db_queries.bal
@@ -2446,6 +2446,25 @@ isolated function deleteEmployeeEmergencyContactsAuditQuery(int personalInfoId) 
 isolated function deletePersonalInfoAuditQuery(int personalInfoId) returns sql:ParameterizedQuery =>
     `DELETE FROM personal_info_audit WHERE personal_info_pk_id = ${personalInfoId};`;
 
+# Resolve the person behind a work email.
+#
+# A person may hold several employee rows over time (a rehire issues a new employee ID while
+# reusing the same personal_info row), so this deliberately returns the personal_info_id rather
+# than an employee ID. LIMIT 1 is safe because every employee row for one person points at the
+# same personal_info_id, and a work email belongs to exactly one employee row.
+#
+# + workEmail - Work email of the person
+# + return - Parameterized query returning the person's personal_info_id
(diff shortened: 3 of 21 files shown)


shayanmalinda:
🤖 **Claude review complete** — 11 comment(s) posted ([view review](https://github.com/wso2-open-operations/people-ops-suite/pull/325#pullrequestreview-4934350913))

🔴 High: 1 · 🟠 Medium: 5 · 🟡 Minor: 5

shayanmalinda (commented):
Read through the whole change: the diff engine and audit queries, the new `promotion` module, the history endpoint and its authorization, the webapp slice and timeline component, and the two continuous-service query fixes. Left 11 comments — 1 high, 5 medium, 5 minor.

Start with the authorization one on `service.bal`. The lead check is scoped to the requested employment row, but the response fans out to every employment the person has ever held, so a former manager who is still recorded on an old row can reach the current one.

Two notes that aren't inline. The PR description lists the CSV external-designation fix as part of this change, but `externalDesignation` is already in `EMPLOYEE_CSV_COLUMNS` on `main` — nothing here touches it. The continuous-service fix, on the other hand, does land in both query sites, and the recursive chain walk looks right.

apps/people-app/backend/service.bal
```diff
@@ -2875,4 +2876,143 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+        // - Self gets the employee projection: no actionBy, no system rows.
+        boolean hasFullProjection = authorization:checkPermissions(
+                [authorization:authorizedRoles.ADMIN_ROLE], userInfo.groups);
+
+        if !hasFullProjection {
+            // Deliberately checked against the requested employeeId itself, never resolved
+            // through personal_info_id first. The lead relationship is carried by manager_email
+            // on an individual employment row, so resolving to the person would let a lead reach
+            // a former employment period of someone who is no longer their report.
+            boolean|error isSubordinate = database:isSubordinateOfLead(userInfo.email, employeeId);
```
shayanmalinda, line 2920: **🔴 High: The lead check passes on a stale employment row, but the response covers the whole person**

`isSubordinateOfLead` matches on `manager_email` of the requested employment row, and that column is never cleared on an old row, so a former manager can request someone's previous employee ID and get back every employment period for that person — including the current one, with `actionBy` attribution. The comment above says resolving through `personal_info_id` would allow exactly this, but the fan-out happens later in `getEmploymentPeriods` regardless. Either restrict the lead projection to the requested employment row, or require that row to be the person's current employment before granting the full projection.

---

apps/people-app/backend/modules/database/db_functions.bal
```diff
@@ -1513,3 +1513,89 @@ public isolated function hasActiveEmployeesInBUTeamSubTeamUnitMapping(int id) re
...
+# + employeePkIds - Employee table primary keys belonging to the person
+# + return - Audit snapshots from employee_audit, personal_info_audit, and
+# employee_additional_managers_audit, ordered by action_on ascending across all three
+public isolated function getAuditSnapshots(int[] employeePkIds) returns AuditSnapshot[]|error {
+    stream<AuditSnapshot, error?> employeeAuditStream =
+        databaseClient->query(getEmployeeAuditSnapshotsQuery(employeePkIds));
+    AuditSnapshot[] employeeAuditSnapshots = check from AuditSnapshot snapshot in employeeAuditStream
+        select snapshot;
+
+    stream<AuditSnapshot, error?> personalInfoAuditStream =
```
shayanmalinda, line 1564: **🟠 Medium: Two of the three audit queries produce no events**

`buildHistoryEvents` only diffs snapshots whose `sourceTable` is `employee_audit`, so the `personal_info_audit` and `employee_additional_managers_audit` rows are fetched, merged and re-sorted for nothing. That is two extra round trips plus their full JSON payloads on every history request — drop both queries until a tracked field actually lives in those tables.

> 📖 **Reference ([dead-code](https://refactoring.guru/smells/dead-code)):** "Unused functions, exports, params, branches, or orphaned writes (a key written but never read). Delete it."

---

apps/people-app/backend/utils.bal
```diff
@@ -742,3 +743,116 @@ public isolated function parseCsvBytes(byte[] fileBytes) returns BulkEmployeeCsv
...
+            previousValue: event.previousValue,
+            currentValue: event.currentValue,
+            occurredOn: event.occurredOn,
+            isSystem: false
+        };
+}
+
+# Attach each promotion to the employment period it took effect during.
+#
+# A promotion belongs to the period whose startDate..endDate range contains its promotedDate.
```
shayanmalinda: **🟠 Medium: The doc comment describes behaviour this function doesn't have**

It says promotions are matched on `promotedDate` and that one falling outside every range is attached to the nearest earlier period, but the code matches on `createdOn` and drops the unmatched one with a warning. Someone chasing a missing promotion will trust the comment first — rewrite it around the `createdOn` containment rule and say the promotion is omitted.

> 📖 **Reference ([comments](https://refactoring.guru/smells/comments)):** "Comments restating the code instead of explaining *why*."

---

apps/people-app/backend/service.bal
```diff
@@ -2875,4 +2876,143 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+            log:printError("Error resolving history lookup names; returning raw ids",
+                    lookupNames, employeeId = employeeId);
+        }
+
+        // The HRIS promotion database is a second database. An outage there must degrade the
+        // timeline rather than break the employee's own profile page, so the failure is
+        // reported as a flag alongside an otherwise complete response.
+        boolean promotionsUnavailable = false;
+        promotion:PromotionRecord[] promotions = [];
+        string workEmail = periods[0].workEmail;
```
shayanmalinda: **🟠 Medium: Promotions are looked up for one work email only**

`getEmploymentPeriods` gathers employment rows by `personal_info_id`, so the periods can legitimately carry different work emails, but promotions are fetched for `periods[0].workEmail` alone. Any promotion raised under an earlier address is silently missing — collect the distinct emails across `periods` and query HRIS for all of them.

---

apps/people-app/backend/Config.toml.local
```diff
@@ -14,10 +14,22 @@ EXTERNAL_USER_ROLES = [];
     maxOpenConnections = 
     maxConnectionLifeTime = 
     minIdleConnections= 
-    [people.database.dbConfig.options]
-    connectTimeout = 
-    [people.database.dbConfig.options.ssl]
-    mode = "PREFERRED"
+
+
+[people.promotion.promotionDbConfig]
```
shayanmalinda, line 22: **🟠 Medium: This hunk deletes the main database's `options` block**

`[people.database.dbConfig.options]` with `connectTimeout`, and `[people.database.dbConfig.options.ssl]` with `mode = "PREFERRED"`, are removed here, which looks unrelated to adding the HRIS section. Anyone setting up locally from this template loses both — restore the block above the new `[people.promotion.promotionDbConfig]` section.

---

apps/people-app/webapp/src/slices/employeeSlice/employeeHistory.ts
```diff
@@ -0,0 +1,150 @@
...
+      return rejectWithValue(errorMessage);
+    }
+  },
+);
+
+const EmployeeHistorySlice = createSlice({
+  name: "employeeHistory",
+  initialState: employeeHistoryInitialState,
+  reducers: {
+    resetEmployeeHistory(state) {
```
shayanmalinda, line 124: **🟠 Medium: `resetEmployeeHistory` is never dispatched, so history goes stale across employees**

Nothing in the webapp calls this reducer, and the slice is global, so it keeps the last employee's timeline. Opening History on a second employee renders the first one's rows until the fetch resolves — either dispatch it from `EmployeeHistory`'s effect cleanup, or clear `state.history` in the `pending` case.

> 📖 **Reference ([dead-code](https://refactoring.guru/smells/dead-code)):** "Unused functions, exports, params, branches, or orphaned writes (a key written but never read). Delete it."

---

apps/people-app/webapp/src/view/me/index.tsx
```diff
@@ -1086,7 +1092,6 @@ export default function Me({
         </AccordionDetails>
       </Accordion>
       <Accordion
```
shayanmalinda, line 1094: **🟡 Minor: The Personal Information accordion no longer opens by default**

`defaultExpanded` is dropped from this accordion, which the History section doesn't need and the PR description doesn't mention. Looks accidental — put it back unless the collapse was intended.

---

apps/people-app/backend/modules/promotion/functions.bal
```diff
@@ -0,0 +1,48 @@
...
+            cycleName: promotion.cycleName,
+            businessUnit: promotion.businessUnit,
+            department: promotion.department,
+            team: promotion.team,
+            subTeam: promotion.subTeam,
+            // createdOn comes from a TIMESTAMP column via DATE(), so it is always
+            // YYYY-MM-DD and needs no normalisation.
+            createdOn: promotion.createdOn,
+            promotedDate: promotion.promotedDate is string
+                ? re `/`.replaceAll(<string>promotion.promotedDate, "-")
```
shayanmalinda: **🟡 Minor: The cast goes away if the field is bound to a local first**

Narrowing doesn't apply to record field access, which is why `<string>` is needed here. Bind it first — `string? promotedDate = promotion.promotedDate;` — and the `is string` branch narrows on its own.

> 📖 **Reference ([avoid_unnecessary_type_casts](https://learn-ballerina.github.io/best_practices/avoid_unnecessary_type_casts.html)):** "Avoid unnecessary type casts"

---

apps/people-app/backend/modules/promotion/types.bal
```diff
@@ -0,0 +1,62 @@
...
+    # Database password
+    string password;
+    # Database name
+    string database;
+    # Database host
+    string host;
+    # Port number of the MySQL server
+    int port;
+    # Connection pool configuration
+    sql:ConnectionPool connectionPool?;
```
shayanmalinda, line 37: **🟡 Minor: No `options` field, so the HRIS connection can't be given an SSL mode**

`DatabaseConfig` in the database module carries `mysql:Options options?`; this record doesn't, so a cross-database connection to HRIS can't require TLS or set a connect timeout. Add `mysql:Options options?` alongside `connectionPool`.

---

apps/people-app/backend/modules/promotion/client.bal
```diff
@@ -0,0 +1,25 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+import ballerinax/mysql;
+import ballerinax/mysql.driver as _;
+
+# HRIS promotion database client configuration.
+configurable PromotionDatabaseConfig promotionDbConfig = ?;
```
shayanmalinda, line 20: **🟡 Minor: `apps/people-app/CLAUDE.md` still says six backend modules**

It lists and describes six, and `promotion` makes seven. Add a line for it next to the `qr` and `scim` entries so the module list stays accurate.

---

docs/specs/2026-08-11-employee-history-plan.md
```diff
@@ -0,0 +1,599 @@
+# Employee History Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
```
shayanmalinda, line 3: **🟡 Minor: Does the build plan belong in the repo?**

This is a 599-line task-by-task plan addressed to agentic workers, and it stops being true the moment the feature merges. The design doc next to it reads as durable documentation — worth keeping that one and leaving the plan (and the mockup HTML) out of the tree.

---

apps/people-app/backend/service.bal
sacheeramesh: Fixed in `cfa92d53`. This was a real hole and the analysis was exactly right — the check guarded the requested row, but the response fanned out to the whole person regardless, so the guard protected the wrong thing.

A lead now needs both conditions: the person is their subordinate **and** the requested row is that person's current employment.

Worth recording that the first attempt at the currency check was wrong. I wrote it as "is this the latest row for this `personal_info_id`", and against the real three-employment record all three periods returned true — because those employments hold three *different* `personal_info_id` values (one NIC recorded three ways), so each was trivially "latest" within its own group of one. It would have passed for every stale row and fixed nothing.

The committed version walks the `continuous_service_record` chain instead, the same one the history query uses. Verified against that record: only the current employment passes, both former ones are denied.

apps/people-app/backend/service.bal
sacheeramesh: Fixed in `cfa92d53`. Confirmed real: `work_email` is a column on the employment row, and a chain is walked by `continuous_service_record` rather than by email, so the periods genuinely can carry different addresses.

Now collects the distinct non-empty emails across all periods and queries each, merging the results. If any single lookup fails, `promotionsUnavailable` is still set, so a partial result is flagged rather than presented as complete.

No employee in the current data has divergent emails across a chain, which is why nothing was visibly broken — but the schema permits it and the job-info update lets an admin change the email on one row.

apps/people-app/backend/service.bal
```diff
@@ -2875,4 +2876,143 @@ service http:InterceptableService / on new http:Listener(9090) {
...
+        promotion:PromotionRecord[]|error fetchedPromotions = promotion:getApprovedPromotions(workEmail);
```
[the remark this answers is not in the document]
sacheeramesh: Fixed in `cfa92d53` — see the reply on the same finding at service.bal:3000 for detail. Distinct emails across all periods are now queried and merged.

apps/people-app/backend/utils.bal
sacheeramesh: Fixed in `e9ce28e8`. The comment described the original design — matching on `promotedDate`, with an unmatched promotion attached to the nearest earlier period. Both changed when attribution moved to `createdOn`, and the comment did not follow.

Rewritten around the `createdOn` containment rule, stating that a non-matching promotion is logged and omitted, and explaining why the promoted date is not used: HRIS frequently leaves it null, whereas a request is always raised while the employee is employed.

apps/people-app/backend/utils.bal
```diff
@@ -742,3 +743,116 @@ public isolated function parseCsvBytes(byte[] fileBytes) returns BulkEmployeeCsv
...
+# An open period (no endDate) extends to the present.
+#
+# A promotion that falls in no range is attached to the nearest *earlier* period rather than
+# dropped, because the HRIS promoted date and the employment start date are maintained in two
+# different systems and disagree at the edges. Losing a real promotion is worse than showing it
+# one period early. A promotion earlier than every period has no earlier period to fall back on
+# and is left unattached.
```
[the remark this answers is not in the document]
sacheeramesh: Fixed in `e9ce28e8` — the comment now describes the `createdOn` containment rule and the omit-with-warning behaviour.

apps/people-app/backend/utils.bal
```diff
@@ -742,3 +743,116 @@ public isolated function parseCsvBytes(byte[] fileBytes) returns BulkEmployeeCsv
...
+#
+# + periods - Employment periods for the person, newest first
+# + promotions - Approved promotions for the person
+# + return - Periods in their original order, each carrying its promotions
```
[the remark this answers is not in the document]
sacheeramesh: Fixed in `e9ce28e8`. Agreed the doc was the stale part rather than the code.

docs/specs/assets/2026-08-11-employee-history-mockup.html
```diff
@@ -0,0 +1,417 @@
+<title>Employee History — People App</title>
```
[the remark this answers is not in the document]
sacheeramesh: Fixed in `e9ce28e8`. The file was authored for a publisher that supplies its own `<!doctype html>` wrapper, so it had none as a standalone file. Added, along with a charset declaration.

apps/people-app/backend/Config.toml.local
```diff
@@ -14,10 +14,22 @@ EXTERNAL_USER_ROLES = [];
     maxOpenConnections = 
     maxConnectionLifeTime = 
     minIdleConnections= 
-    [people.database.dbConfig.options]
-    connectTimeout = 
-    [people.database.dbConfig.options.ssl]
-    mode = "PREFERRED"
+
+
```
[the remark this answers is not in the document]
sacheeramesh: Fixed in `5b0ec25e`. Entirely accidental — adding the HRIS section deleted `[people.database.dbConfig.options]` along with `connectTimeout` and the SSL mode. Restored, and the new HRIS block now carries the same placeholders.

apps/people-app/backend/Config.toml.local
sacheeramesh: Fixed in `5b0ec25e`. Restored `connectTimeout` and `mode = "PREFERRED"` for the primary database. The deletion was accidental and unrelated to this feature.

apps/people-app/backend/modules/promotion/types.bal
sacheeramesh: Fixed in `5b0ec25e`. Added `mysql:Options options?` to `PromotionDatabaseConfig`, with matching placeholders in `Config.toml.local`. Worth having given this is a cross-machine connection to an HR database, and the primary config already supported it.

apps/people-app/backend/modules/promotion/client.bal
sacheeramesh: Fixed in `5b0ec25e`. `CLAUDE.md` now says seven modules and describes `promotion`, including why it is separate from `database`: a second datasource with its own client, config and types, so the dependency can be replaced without touching the primary module.

apps/people-app/backend/modules/promotion/functions.bal
sacheeramesh: Fixed in `5b0ec25e`. Bound `promotedDate` to a local in the query expression so narrowing applies and the cast is gone.

apps/people-app/webapp/src/slices/employeeSlice/employeeHistory.ts
sacheeramesh: Fixed in `5b0ec25e`, and slightly more than suggested. `resetEmployeeHistory` is now dispatched on unmount **and** the thunk clears `history` in its `pending` case.

The `pending` clear is the one that actually closes the hole — it covers every entry point regardless of how the fetch was triggered — while the unmount cleanup stops a stale timeline sitting in the store between visits. Either on its own leaves a gap.

apps/people-app/webapp/src/slices/employeeSlice/employeeHistory.ts
```diff
@@ -0,0 +1,150 @@
...
+  events: HistoryEvent[];
+  promotionsUnavailable: boolean;
+}
+
+interface EmployeeHistoryState {
+  state: State;
+  stateMessage: string | null;
+  errorMessage: string | null;
+  history: EmployeeHistoryResponse | null;
+}
```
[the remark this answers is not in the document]
sacheeramesh: Addressed in `5b0ec25e`, though by a different mechanism than proposed. Rather than tracking `requestId` and the active employee on `pending`, the thunk now clears `history` there and the component dispatches `resetEmployeeHistory` on unmount.

That removes the stale-render window without adding request-identity bookkeeping. The overlap this guards against also needs two histories in flight at once, and the component fetches once on mount after the user expands the section — navigating to another employee unmounts it, so the second fetch is a fresh mount rather than a concurrent one.

apps/people-app/webapp/src/view/me/index.tsx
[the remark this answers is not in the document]
sacheeramesh: Intentional, not accidental — but the PR description did not say so, which is why it reads as a stray change. That is a gap in the description rather than in the code.

The profile page was opening with two expanded sections and a third added by this PR. Collapsing Personal Information keeps the default page short, with General Information as the only section open. The personal-info fetch still runs on mount rather than on expand, so there is no spinner when the section is opened.

sacheeramesh: Deliberate and requested, though the PR description omits it — flagging it was the right call on the evidence available.

Rationale: this PR adds a third section to the profile, and leaving two expanded plus a new one made the default page long. General Information stays open; Personal Information and History are collapsed. The personal-info fetch is unchanged and still happens on mount, so opening the section shows data immediately.

apps/people-app/backend/modules/database/history.bal
```diff
@@ -0,0 +1,185 @@
...
+# 2. Snapshots are compared within a single employee record. A person may hold several
+# employee rows over time (a rehire), and their snapshots interleave in one time-ordered
+# stream. Comparing across records would invent transitions that never happened.
+#
+# 3. Snapshots are compared within a single source table. The three audit tables carry
+# different columns under partly overlapping names, so a cross-table comparison would
+# report tracked fields flipping to null purely because the other table lacks them.
+#
+# Only TRACKED_EMPLOYEE_FIELDS produce events; all of them live in employee_audit, so
+# snapshots from the other audit tables establish baselines but currently emit nothing.
```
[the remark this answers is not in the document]
sacheeramesh: Partly done since this was written. `personal_info_audit` now has its own tracked field list and produces events — all 17 real columns, excluding only the generated `full_name`, which would otherwise emit a duplicate event for every name change. Events carry their `sourceTable`, and the timeline groups them into Employment / Personal Info / Promotions views.

`employee_additional_managers_audit` still emits nothing. It is left as-is deliberately: the primary manager is already tracked through `manager_email` on the employee row, so what remains is *additional* managers, and a stable relationship key for those is a separate design question rather than a missing field list.

apps/people-app/backend/modules/database/db_functions.bal
sacheeramesh: Half of this is now obsolete: `personal_info_audit` produces events as of `f30ace1a`, so that query is no longer a wasted round trip.

`employee_additional_managers_audit` is still fetched for nothing, and the point stands for it. Leaving it rather than deleting it, on the grounds that the primary manager is already covered by `manager_email` and additional-manager history needs a stable relationship key before it can be diffed meaningfully — I would rather that arrive as a considered change than remove and re-add the query. Happy to drop it if you would prefer the tree clean in the meantime.

apps/people-app/backend/modules/database/db_queries.bal
```diff
@@ -220,7 +231,18 @@ isolated function getEmployeesQuery(EmployeeSearchPayload payload, string? leadE
...
+            (
+                WITH RECURSIVE chain AS (
+                    SELECT csr0.id, csr0.continuous_service_record, csr0.start_date
+                    FROM employee csr0 WHERE csr0.id = e.continuous_service_record
+                    UNION
+                    SELECT csr1.id, csr1.continuous_service_record, csr1.start_date
+                    FROM employee csr1 JOIN chain ON csr1.id = chain.continuous_service_record
+                )
+                SELECT MIN(chain.start_date) FROM chain
+            ) AS continuousServiceDate,
```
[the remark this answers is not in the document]
sacheeramesh: Fair, and not measured either way — so I would rather flag it than claim it is fine.

The correlated subquery was chosen because it fixed a correctness bug: the previous single-hop `LEFT JOIN` returned the wrong continuous service date for anyone with three or more linked employments, understating their service by years. A set-based recursive CTE joined once would give the same answer more cheaply on the list paths.

Not doing it in this PR: the wrong-date fix is already scope beyond the feature, and swapping the shape without measuring would trade a known-correct query for an unmeasured one. Worth its own change with timings from a realistic export, and I am happy to raise it as a follow-up unless you would rather see it here.

docs/specs/2026-08-11-employee-history-plan.md
sacheeramesh: Agreed, and I raised the same reservation when these docs were first added: the design doc is durable, the plan is scaffolding addressed to whoever builds the feature and stops being true the moment it merges.

Left in for now only because including it was a deliberate call rather than an oversight, so removing it is the author’s decision to reverse rather than mine to assume. If the preference is that specs stay and plans do not, I will drop the plan and the mockup and keep the design doc.

v15a1 (approved):

PR #326, opened 2026-08-14
[People App] Maintainance Banner

NimsaraRavindu:
#### 📝 Summary
This PR introduces a global **Maintenance Mode** switch for the People App. It allows administrators to toggle a configuration flag in the backend that will instantly render a full-screen, blocking maintenance banner on the frontend, preventing users from interacting with the app during scheduled downtime.

---

##### Frontend (`microapp`)
* **Maintenance Banner Component:** Created a new `<MaintenanceBanner />` component that acts as a `z-[9999]` fixed overlay with a blurred backdrop, blocking all clicks to the underlying UI.
* **Theme Consistency:** The banner is fully responsive and supports both Light and Dark modes seamlessly, strictly adhering to the app's brand colors (e.g., `#1F2A44`, `#ff7300`, `#808080`).
* **Global Injection:** Rendered the banner inside the root `HashRouter` in `App.tsx` so it overlays across all pages and routes.
* **Config Fetching:** Added a new `fetchAppConfigs` endpoint definition to `config.ts`. The banner polls this on mount to determine its visibility state.

##### Backend (`backend`)
* **Config API Endpoint:** Added a new public `/configs` resource function in `service.bal` that returns global application configurations.
* **Configurable Record:** Defined a new `AppConfig` Ballerina record containing the `isMaintenanceMode` boolean, allowing zero-downtime configuration via `Config.toml`.
* **JWT Interceptor Bypass:** Updated `authorization.bal` to bypass JWT verification exclusively for the `/configs` route. This is required so the frontend can read the maintenance status before a user logs in.
* **Security Hardening:** Removed local-only testing hacks (such as the `Authorization: Bearer` manual fallback) from the `JwtInterceptor` to ensure the codebase remains strictly production-ready.

---

#### 🧪 Testing Instructions

1. Run the backend and frontend locally.
2. In `Config.toml.local`, add the following:
   ```toml
   [people.app]
   isMaintenanceMode = true

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **New Features**
  * Added configurable maintenance mode for the People app.
  * Displays a full-screen notice when the app is temporarily unavailable.
  * Added endpoints for retrieving application configuration and user information.

* **Bug Fixes**
  * Configuration requests can now be accessed without JWT authorization, ensuring maintenance status loads correctly.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->


apps/people-app/microapp/src/components/MaintenanceBanner.tsx
```diff
@@ -0,0 +1,40 @@
+import { useEffect, useState } from "react";
```
shayanmalinda, line 17: Let's add the license headers.

apps/people-app/backend/modules/authorization/authorization.bal
```diff
@@ -27,6 +27,10 @@ public isolated service class JwtInterceptor {
     isolated resource function default [string... path](http:RequestContext ctx, http:Request req)
         returns http:NextService|http:Forbidden|http:InternalServerError|error? {
 
+        if path.length() > 0 && path[0] == "configs" {
+            return ctx.next();
+        }
+
```
shayanmalinda: Do we need to whitelist this API without enforcing Authentication? I don't think so.

apps/people-app/microapp/src/components/MaintenanceBanner.tsx
NimsaraRavindu: done

apps/people-app/backend/service.bal
```diff
@@ -32,6 +32,12 @@ import ballerina/time;
     id: "people-ops-suite/people-service"
 }
 
+public type AppConfig record {|
+    boolean isMaintenanceMode = false;
+|};
+
+configurable AppConfig app = ?;
```
shayanmalinda: We don't need a separate configuration record for this. Let's simply keep a flat boolean record.

apps/people-app/backend/modules/authorization/authorization.bal
NimsaraRavindu: I have removed the whitelist bypass here so that /configs strictly enforces JWT authentication like the rest of the endpoints. I've also updated the <MaintenanceBanner /> on the frontend to use our authenticated 
useHttp client instead of a raw fetch so it passes the token correctly. Included in the latest commit

apps/people-app/backend/service.bal
NimsaraRavindu: done

apps/people-app/backend/service.bal
```diff
@@ -64,6 +65,13 @@ service http:InterceptableService / on new http:Listener(9090) {
     public function createInterceptors() returns http:Interceptor[] =>
         [new authorization:JwtInterceptor(), new ErrorInterceptor()];
 
+    # Get global app config.
+    #
+    # + return - App config
+    resource function get configs() returns AppConfig {
+        return app;
+    }
+
```
shayanmalinda, line 74: Do we need a seperate new endpoint? Don't we have an existing endpoint which is being invoked when navigated to Car Park Booking UI?

shayanmalinda: Let's check the possibility of using that.

apps/people-app/microapp/src/config/config.ts
```diff
@@ -36,6 +36,8 @@ export const IS_MICROAPP = window.config.IS_MICROAPP;
 const baseUrl = window.config.BACKEND_BASE_URL;
 
 export const serviceUrls = {
+  fetchAppConfigs: () => `${baseUrl}/configs`,
+  fetchUser: () => `${baseUrl}/user-info`,
```
shayanmalinda, line 40: Do we need this?

apps/people-app/backend/service.bal
NimsaraRavindu: Even when maintenance mode is off, <MaintenanceBanner/> continuously polls every 30s in the background to detect changes.

If we hook into /parkings/configs, every active user across the entire app  would be hitting the parking endpoint every 30 seconds. A dedicated /configs endpoint avoids putting that continuous global traffic on the parking module and keeps concerns cleanly separated.

apps/people-app/microapp/src/config/config.ts
NimsaraRavindu: Yes, this is required to call the generic /configs endpoint.Please see my reply on the service.bal thread above regarding why we shouldn't couple this global banner to the parkings API

apps/people-app/backend/service.bal
shayanmalinda: Agreed

shayanmalinda (approved):

